PR #1, opened 2019-08-26
Moving changes for remotable vivado to api-generator-sifive

sistillson:
Made changes into api-generator-sifive to enable vivado to run as a non-local job.


top.wake
```diff
@@ -34,7 +34,14 @@ global def bitstreamTarget board dut buildRoot =
 
   def topModule = dut.getDUTTopModule
 
-  makeBitstreamPlan name vendor board topModule buildRoot.getBuildTreeMCSDir
+  def visibleFiles =
+    def firrtlRoot = getBuildTreeFirrtlDir buildRoot
+    sources firrtlRoot `.*\.sdc`
+    ++ sources firrtlRoot `.*\.sdc`
```
richardxia: These two lines are duplicates of each other.

top.wake
```diff
@@ -34,7 +34,14 @@ global def bitstreamTarget board dut buildRoot =
 
   def topModule = dut.getDUTTopModule
 
-  makeBitstreamPlan name vendor board topModule buildRoot.getBuildTreeMCSDir
+  def visibleFiles =
+    def firrtlRoot = getBuildTreeFirrtlDir buildRoot
+    sources firrtlRoot `.*\.sdc`
```
richardxia: You're not supposed to use `sources` for getting the paths to generated files. `sources` is meant to only describe "source" files, i.e. files that are committed to source control. I believe the next release of `wake` will enforce this. Instead, I believe the correct way of getting `Path` objects corresponding to generated files is to get them directly from the Job outputs.

top.wake
```diff
@@ -34,7 +34,14 @@ global def bitstreamTarget board dut buildRoot =
...
+    ++ sources firrtlRoot `.*\.sdc`
+    ++ sources firrtlRoot `.*\.json`
```
richardxia: Which JSON files are actually being used? The `.json` extension is not a great one to be filtering by, since there are a lot of files with very different semantic purposes that use that extension. I'm not sure I can remember off the top of my head any JSON files that actually should be used by the Bitstream building jobs.

fpga/mcs.wake
```diff
@@ -8,10 +8,11 @@ tuple BitstreamPlan =
   global Vsrcs:     List Path
   global Resources: List String
   global UseDeputy: Boolean
+  global VisibleFiles: List Path
 
-global def makeBitstreamPlan name vendor board topModule outputDir =
+global def makeBitstreamPlan name vendor board topModule outputDir visibleFiles =
```
richardxia: One comment I want make about this is that this is making a backwards incompatible change to a globally-defined function. Something we need to get better about as we split up the codebase is making sure that we do a better job of maintaining stable interfaces.

richardxia (changes requested):

top.wake
sistillson: fixed

top.wake
sistillson: fixed

top.wake
sistillson: will check

fpga/mcs.wake
sistillson: agreed, but this is the best place to do it...

richardxia: Have we started tagging this repo with versions (such as sem-ver)? If so, we should at least increment part of the version number to indicate that a breaking change has been made.

albertchen-sifive: no we have not tagged this repo. and shouldn't this just use `Nil` as the default for `visibleFiles` instead of adding a new argument to this function?

albertchen-sifive (approved):

PR #4, opened 2019-10-01
Fram-176 Wake workflow for modular Freedom Metal drivers

sistillson:
Initial revision of fram-176.

Creates basic driver code based on the object model and the duh file. Creation based on the
wake flow. Not ready, waiting on wake-language-sifive.


scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+            return '\n'.join(line.strip() for line in long_str.splitlines() if line)
+
+        usage = f"""
+        {sys.argv[0]} <object model file> \
+        <company name> <device name> <metal directory>
+        """
+
+        print("Not enough arguments", file=sys.stderr)
+        print(f'{clear_arg_str(usage)}', file=sys.stderr)
+        sys.exit(1)
```
richardxia: Is there a reason you didn't use `argparse` for argument parsing? It would handle most of this automatically, not to mention it would also provide functionality for easily adding help text and type casting.

I would also ask that we make these flags and not positional arguments, since there's four arguments and there isn't an obvious, natural ordering to them.

dut/drivers.wake
```diff
@@ -0,0 +1,59 @@
+
+tuple SkeletalDriverPlan =
+  global DuhFile:             Path
+  global CompanyName:         String
```
richardxia: Vendor is a more appropriate name than Company here, since not all vendors are companies, and the word vendor makes clearer the role of providing the IP. This would also be more consistent with every other place that we have that uses the word "vendor".

dut/drivers.wake
```diff
@@ -0,0 +1,59 @@
+
+tuple SkeletalDriverPlan =
```
richardxia: What does the word "Skeletal" signify here? Does this code only apply to the Test Socket, which was formerly referred to as the Skeleton? Can we make this code more general and apply to designs beyond the Test Socket?

dut/drivers.wake
```diff
@@ -0,0 +1,59 @@
...
+  global DuhFile:             Path
+  global CompanyName:         String
+  global DeviceName:          String
+  global LocalMetalDirectory: String
+  global DUTPlan:             RocketChipDUTPlan
+
+global def makeSkeletalDriverPlan = SkeletalDriverPlan _ _ _ _ _
```
albertchen-sifive: nitpick: we should try to use named arguments for global functions like
```
global def makeSkeletalDriverPlan duhFile company device metalDir dut =
  SkeletalDriverPlan duhFile company device metalDir dut
```

it makes it easier to read and also helps wake output better type information when you do `wake -v makeSkeletalDriverPlan`

dut/rocket-chip-dut.wake
```diff
@@ -5,7 +5,7 @@ tuple ScalaBlock =
 global def makeScalaBlock module config =
   ScalaBlock module config
 
-tuple RocketChipDUTPlan =
+global tuple RocketChipDUTPlan =
```
albertchen-sifive: why is this global? you aren't actually calling the constructor, so this should remain private.

dut/drivers.wake
```diff
@@ -0,0 +1,59 @@
...
+    then Some aPlan
+    else None
+  mapPartial samePlan plans
+  | head
+
+def skeletalPlanToPlan skeletalDriverPlan =
+  def objModel =
+    makeRTL (skeletalDriverPlan.getSkeletalDriverPlanDUTPlan)
+    | getDUTObjectModel
+    | prettyJSON
```
albertchen-sifive: why not just put an `ObjectModel` field in the `SkeletalDriverPlan`? The `DUTPlan` has many fields you shouldn't care about, and you need to run the whole generator flow to get object model from it.

dut/drivers.wake
sistillson: Fixed

dut/rocket-chip-dut.wake
sistillson: I'm using it for typing in a tuple

dut/drivers.wake
sistillson: If you have already generated the rtl, it's pretty fast. If you haven't generated the RTL you wouldn't have it and you would have to generate the RTL. I'm not sure how we would expect the Object model to come from besides makeRTL.

albertchen-sifive: but this is running a big job that really isn't this function's job to run. what if the `DUTPlan` passed in isn't actually a config that the user want to be built? or if someone needs to generate drivers for an existing object model, but they don't have access to the `DUTPlan` used to make that object model?

sistillson: Then where is the object model coming from?  If the code isn't tied to the dut, what is the point? Do we just ask them questions like duh?

albertchen-sifive: Just assume whoever is calling this function has an object model. If the caller does not have one then they will call `makeRTL` or whatever. The code shouldn't be tied to the dut. it should only be tied to the object model.

dut/rocket-chip-dut.wake
richardxia: I don't think you need to use the `global` keyword for using it in a type, even if that is kind of confusing. `global` here only refers to globally exporting the constructor.

See some of the discussion in Wake about how we want to handle namespacing and specifically regarding type names (as opposed to constructors).

dut/drivers.wake
```diff
@@ -0,0 +1,59 @@
...
+
+  makePlan cmd visibleFiles
+
+def makeSkeletalDrivers skeletalDriverPlan =
+  skeletalPlanToPlan skeletalDriverPlan
+  | addPythonEnv ("{here}/../scripts/om2files_env".simplify)
+  | runJob
+  | getJobStderr
+
+global def createSkeletalDriver deviceName =
```
richardxia: For all `global` functions can we please write a comment document what it does, what it takes in, and what it returns?

I'm also kind of confused by the return type here. It looks like it's an `Option (Result String Error)`, where the `Some (Pass String)` case is the standard output text. What is a caller meant to do with the return type? Shouldn't we make sure that the return type actually includes Paths to the generated driver files?

dut/drivers.wake
```diff
@@ -0,0 +1,59 @@
...
+    | prettyJSON
+
+  def name =
+    skeletalDriverPlan
+    | getSkeletalDriverPlanDUTPlan
+    | getRocketChipDUTPlanName
+
+  def omfile = write "{workspace}/build/api-generator-sifive/{name}/objectmodel.json" objModel
+  def cmd =
+    (which "python3.7"),
```
richardxia: We shouldn't be using `which`. My understanding is that this will bypass the runner system by evaluating the path to the `python3.7` executable prior to handing it off to runner.

dut/drivers.wake
richardxia: I second @albertchen-sifive's suggestion about having this function/Plan take in a path to an object model JSON file as a parameter and not construct the object model itself.

dut/drivers.wake
```diff
@@ -0,0 +1,59 @@
...
+  global DuhFile:             Path
+  global CompanyName:         String
+  global DeviceName:          String
+  global LocalMetalDirectory: String
+  global DUTPlan:             RocketChipDUTPlan
+
+global def makeSkeletalDriverPlan = SkeletalDriverPlan _ _ _ _ _
+
+def getDevicePlan deviceName =
+  def plans = subscribe skeletalDriverPlans
```
richardxia: Can you document the `skeletalDriverPlans` pub/sub topic? This is an important API for users of this flow, since they need to understand what data type should be published and what the expected behavior is of publishing to this topic.

dut/drivers.wake
```diff
@@ -0,0 +1,59 @@
...
+  def samePlan aPlan =
+    if aPlan.getSkeletalDriverPlanDeviceName ==~ deviceName
+    then Some aPlan
+    else None
+  mapPartial samePlan plans
+  | head
```
richardxia: This feels like an awkward construction in Wake, since it obfuscates what I believe is your intent. I think you're just trying to find the first item in a list that matches a predicate, so you should just use the [`find`](https://github.com/sifive/wake/blob/a91f5e84f780fb620adb0c6ad14cf80114c2dc46/share/wake/lib/core/list.wake#L139) function.

richardxia (changes requested):
I haven't taken a look at the om2files.py file yet, but I wanted to leave some comments about the Wake rules first.

dut/rocket-chip-dut.wake
sistillson: fixed

dut/drivers.wake
sistillson: I think you will make the customer code more complicated and prone to error.

But, I will change it.

scripts/om2files.py
sistillson: fixed

dut/drivers.wake
sistillson: fixed

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+import string
+
+from dataclasses import dataclass
+
+###
+# JSON support
+###
+
+# type variable to handle our use of json
+JsonType = t.TypeVar('JsonType', dict, list)
```
richardxia: I'm a little confused by what your exact intent is here with this type. Here, you only included the collection types but not any of the primitives, like strings, numbers, or booleans. However, in several of your functions that take in a  `JsonType` as an argument, you do type checking for some of these primitive cases. Can you either update this type variable to be a full union of all possible JSON values or go 100% the other way and just make this a `TypeVar('JsonType', object)` and not attempt to use it for actual type checking but for humans to read? It's a little bit confusing to have this be halfway in between.

Also, let's follow [PEP 8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles) and capitalize all the letters in "JSON". Most of the rest of the Python standard library follows that convention, so it'll be more consistent to only have one style of capitalizing acronyms.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+
+# Check if file exist before over-writing it
+CHECK_EXISTENCE = True
```
richardxia: Can you make this an argparse argument? It's a little strange to have this be a constant when it otherwise looks like a configurable value.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
+#!/usr/bin/env python3.7
```
richardxia: This isn't a comment about this particular line but about the entire file itself, since I don't have a different place to put it. Can we give this script a better name? `om2files` is not really descriptive because it doesn't say anything about what files it is creating.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+JsonType = t.TypeVar('JsonType', dict, list)
+
+# Check if file exist before over-writing it
+CHECK_EXISTENCE = True
+
+
+@dataclass(frozen=True)
+class Register:
+    """
+    Data class to hold register information
```
richardxia: This docstring doesn't communicate any more information than you would get from reading the class declaration. How about something more like:

> Description of memory-mapped control register within a device.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+    name: str
+    offset: int  # in bytes
+    width: int  # in bits
+
+
+class InvalidOM(Exception):
+    """
+    Raised when an incorrect object is pass as something that should
+    have been json
```
richardxia: I'm not sure I understand what this docstring is saying, as it sounds inconsistent with both the actual name of the class as well as its usage elsewhere in the code. The "OM" in "InvalidOM" presumably stands for "Object Model", but there's no reference to the object model in this docstring. Also, in the code, this exception is thrown when an object is not a `dict` or a `list`, but JSON includes more than just `dict` and `list`.

Can you clarify the intent of this exception is and update its name, its description, and possibly its usage?

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+    pass
+
+
+def search_json_component(js_obj: JsonType, regex: re.Pattern) -> t.Generator:
```
richardxia: It's not helpful to indicate a return type that is a collection but without any description of items within the collection.

Also, Generator is too generic of a type to use here. The Python definition of a Generator includes not only an object that can yield items but an object that can _receive_ input back into it via the `send()` method. You can tell if you have a more general generator by whether you make use of the return value of a `yield` expression, and here you don't ever make use of the return value of `yield` expressions. This is why the `typing.Generator` type accepts three type parameters; one for the yield type, one for the send type, and one for the return type. See [PEP 342](https://www.python.org/dev/peps/pep-0342/) and the [`typing.Generator` docs](https://docs.python.org/3/library/typing.html#typing.Generator) for more details.

For this particular function, you probably want to use `typing.Iterator` or `typing.Iterable` instead.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+    Search a json object for any key or string matching regex. Return a
```
richardxia: This description is inconsistent with the function argument type annotation. You say that it "searches a JSON object", but your code here also explicitly checks the type and handles the cases where you have a list or a string. Note that the term "object" in JSON refers specifically to the key-value mapping type and not the notion of an instance of a class.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    generator for those objects.
+
+    :param js_obj: A parsed javascript object
+    :param regex: a compiled regular expression to search for.
+    :return: A generator returning matching parts
+    """
+    if isinstance(js_obj, list):
+        for i in js_obj:
+            for j in search_json_component(i, regex):
+                if j is not None:
```
richardxia: I'm not understanding when this would ever yield a `None` value. Neither the docstring nor the actual code indicates that `search_json_component()` would ever `None`, so it's a bit confusing to check for that case.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+        for k, v in js_obj.items():
+            if regex.match(k) or \
+                    isinstance(v, str) and regex.match(v):
+                yield js_obj
+                sub_search = False
+                break
+        if sub_search:
+            for d in search_json_component(list(js_obj.values()), regex):
+                if d is not None:
+                    yield d
```
richardxia: This looks like it's unnecessarily constructing a list in order to call `search_json_component()` on it, when you could just directly iterate over the values and recursively call `search_json_component()` on each value.

```py
for v in js_obj.values():
    yield from search_json_component(v, regex)
```

scripts/om2files.py
richardxia: Now that I've fully read the code of this function, I have some questions about the semantics and intended behavior of the function. I see that you yield a match if any string matches the given regex, whether the string is present in an object's keys, an object's values, or a list's items. More precisely, this function yields one of the following:

- A Python dict if either a key matches the regular expression or if value matches the regular expression
- A string if the string matches the regular expression

This means that the type of the yielded value is a `Union[dict, str]`. For a reader of this code, it's kind of difficult to understand the behavior of the function, since it does one of two very different things depending on the inputs, and the return value is a mixed bag of things.

For example, this is one potential return value:

```py
my_dict = {"key": ["foo", "foobar", "bar"], "key2": {"foobar": "foo"}}
list(search_json_component(my_dict, re.compile(r"foo")))
# Returns ['foo', 'foobar', {'foobar': 'foo'}]
```

Is the caller of this function intended to further filter down to the type of thing they actually want?

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+
+    :param js_obj: A dict or list from a parsed json file
+    :param regex_str: A regular expression string to search js_obj for
+    :return: a list of matching parts
+    """
+    regex = re.compile(regex_str)
+
+    if isinstance(js_obj, list) or isinstance(js_obj, dict):
+        return [i for i in search_json_component(js_obj, regex) if i is not None]
```
richardxia: Once again, you do not need to check for `None` values because it's not possible to get that from `search_json_component()`.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    else:
+        raise InvalidOM
```
richardxia: Can you include a message with this exception? Otherwise it'll be pretty frustrating to get this exception and no indication of what was specifically invalid.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+
+
+###
+# templates
+###
+def clean_str(a_str: str) -> str:
+    """
+    Clean up a multiline string so that the line begins after the '|'
+    character. Useful for guaranteeing exact format of a long string.
```
richardxia: This seems like it's duplicating functionality in the Python standard library, specifically [`textwrap.dedent()`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) function. I know that `dedent()` does not use the vertical bars, but it has a similar effect of stripping leading whitespace on each line so that each line still has the same amount of whitespace relative to the first line.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    clean_str(a) => 'this is\n  that is\n\n foo'
+    """
+    def post_pipe(b_str: str):
+        return b_str.split('|', 1)[-1]
+    return "\n".join(post_pipe(l).rstrip() for l in a_str.splitlines())
+
+
+def generate_vtable(device_name: str, reg_list: t.List[Register]) -> str:
+    """
+    Generate the vtable entries for a device and set of register.
```
richardxia:
```suggestion
    Generate the vtable entries for a device and set of registers.
```


Can you also add a description of what the vtable entries are for? Also, are these just the prototypes and not the definitions?

richardxia (commented):
Sorry, still haven't finished reading all of om2files, but wanted to submit my current comments so far.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    :return: thh c code for the vtable entries
+    """
+
+    rv = []
+
+    for a_reg in reg_list:
+        reg_name = a_reg.name.lower()
+        size = a_reg.width
+        if size not in (8, 16, 32, 64):
+            raise Exception('weird size')
```
richardxia: Let's write a more helpful error message. I'd ask that you consider what it would feel like to see this error message as a user running the code. Some things that I think would be helpful to convey in the error message are:

- What the expected value was
- What the received value was
- What this actually corresponds to in the user input

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+        write_func = f'    void (*v_{device_name}_{reg_name}_write)' \
+                     f'(uint32_t * {device_name}_base, uint{size}_t data);'
```
richardxia: I would request that we don't break up the strings into multiple lines for generated code like this, since it makes it more difficult to read.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+
+        rv.append(write_func)
+        rv.append(read_func)
+
+    return '\n'.join(rv)
+
+
+def generate_metal_dev(devices_name: str) -> str:
+    """
+    Generate the metal device for the given device name
```
richardxia: What does it mean to generate the "device"? The "device" is a piece of hardware, so this is presumably a software representation of the device. What kind of representation is it?

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    arg_dict = {'company': company,
+                'dev': dev,
+                'cap_dev': dev.upper(),
+                'index': str(index),
+                'base_address': hex(base_address),
+                'vtable': generate_vtable(dev, reglist),
+                'metal_dev': generate_metal_dev(dev),
+                'protos': generate_protos(dev, reglist)}
+
+    return template.substitute(**arg_dict)
```
richardxia: Nit: I think it'd be clearer and more direct to just pass in the keyword arguments into this method call rather than constructing a dictionary and then splatting it into the method.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+
+# The template for the .h file
+
+METAL_DEV_HDR_TMPL = \
+    """
+    |#include <metal/compiler.h>
+    |#include <stdint.h>
+    |#include <stdlib.h>
+    |
+    |#ifndef ${company}_${dev}${index}_h
```
richardxia: Could we use the word "vendor" instead of "company", which is more correct and more consistent with everything else?

Also, could we use a different name than `dev`? It's not immediately obvious that `dev` stands for device as opposed to other possible things, such as "development". Also, I don't think it would be completely obvious that a variable named "device" holds a string representation of the name of the device as opposed to some more structured information about the device itself. In fact, you have a separate variable named `metal_dev` here which actually holds C code, so it's kind of confusing to see the word `dev` represent two very different aspects of a device.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+    cap_dev = dev.upper()
+    rv: t.List[str] = []
+
+    for a_reg in reg_list:
+        name = a_reg.name.lower()
+        cap_name = a_reg.name.upper()
+        size = a_reg.width
+        if size not in (8, 16, 32, 64):
+            raise Exception('Non-standard size')
```
richardxia: Can we just do this validation once upfront rather than in each of these functions? It would also obviate the issue with having a different exception message for each instance of this check.

dut/drivers.wake
sistillson: I didn't realize that test socket was formerly skeleton. I only used it to mean these driver files are, like a skeleton crew on a ship, a minimal implementation that should be fleshed out. I will replace it with basic, as in these are simple drivers that you might want to expand upon later.

fixed

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+                      file=sys.stderr)
+
+        header_file_path = m_hdr_path / f'{company}_{device}{index}.h'
+        if not header_file_path.exists() and not CHECK_EXISTENCE:
+            header_file_path.write_text(
+                generate_metal_dev_hdr(company, device, index, base, reglist))
+        else:
+            print(f"{str(header_file_path)} exists, not creating.", file=sys.stderr)
+
+        sys.exit(0)
```
richardxia: Is this meant to be here, or was this just a leftover debugging thing? If I'm reading this correctly, it will always terminate right at the end of the first `for` loop iteration.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+
+    devices_enumerated = list(enumerate(devices))
+
+    m_dir_path = Path(metal_dir)
+    m_hdr_path = m_dir_path / device
+
+    if not m_hdr_path.exists():
+        m_hdr_path.mkdir()
+
+    once: bool = True
```
richardxia: Nit: I'd prefer that we don't write type annotations for any local variables that are immediately assigned a literal value, since it's pretty obvious from reading the line that this is a boolean.

scripts/om2files.py
```diff
@@ -0,0 +1,567 @@
...
+
+    for i, om in devices_enumerated:
+        index = i
+        base = om['addressSets'][0]['base']
+
+        if once is True:
```
richardxia: It's not idiomatic in Python to do `is` comparisons with booleans, and you'd normally just write `if once`.

scripts/om2files.py
sistillson: fixed

dut/drivers.wake
sistillson: fixed

dut/drivers.wake
sistillson: fixed

dut/drivers.wake
sistillson: removed

dut/drivers.wake
sistillson: done

dut/drivers.wake
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: renamed to generate_drivers.py

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed. didn't know about that function.

scripts/om2files.py
sistillson: just declartions. will add to name/comment

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

scripts/om2files.py
sistillson: fixed

albertchen-sifive:
I doesn't look like any of the comments on the wake files were addressed even though you marked all the comments as resolved?

sistillson:
My bad. I forgot to git push.

dut/drivers.wake
```diff
@@ -0,0 +1,106 @@
...
+    ++ extraFlags
+  def visibleFiles =
+    "{here}/../scripts/generate_drivers.py".simplify.source,
+    basicDriverPlan.getBasicDriverPlanDuhFile,
+    omfile,
+    basicDriverPlan.getBasicDriverPlanLocalMetalDirectory.source,
+    Nil
+  makePlan cmd visibleFiles
+
+global tuple DriverOutput =
```
albertchen-sifive: the constructor for this tuple should be private since only your driver functions should be able to make these tuples. also we don't want people to pattern match on this tuple in case we want to add more fields in the future.

dut/rocket-chip-dut.wake
albertchen-sifive: this is still global

program/freedom-metal-program.wake
```diff
@@ -179,14 +179,31 @@ global def freedomMetalDUTProgramCompiler =
   def score        = (\_ \_ 1.0) # TODO
 
   def imp dut programCompileOptions =
+    def omFile =
+      dut.getDUTObjectModel
+      | prettyJSON
+      | write "{workspace}/build/api-generator-sifive/{dut.getDUTName}/programs/omfile.json"
```
albertchen-sifive: we should probably add an `ObjectModelFile` field so downstream build steps don't have to write it to another file. but that can be done later

sistillson: I only need to write it to a file to use it in an external script. Eventually this could probably all be rewritten in wake.

dut/drivers.wake
sistillson: fixed

program/freedom-metal-program.wake
sistillson: the other part is fixed too.

albertchen-sifive: also you shouldn't be writing an absolute path. just do `write "build/api-generator-sifive/{dut.getDUTName}/programs/omfile.json"`

sistillson: fixed

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
+
+tuple BasicDriverPlan =
+  global DuhFile:             Path
```
richardxia: Apologies for not commenting on this earlier, but this file is known as the DUH document, which differentiates it from other DUH files. You can see that the [DUH README](https://github.com/sifive/duh) consistently refers to it as a DUH document. Can we rename this to `DUHDocument`?

Also, can we be consistent about how we CamelCase words or other all-caps words? We have both `DuhFile` and `DUTName`, and it makes it more difficult when we have to remember the different and arbitrary capitalization rules. Let's just do all-caps, which seems to be the more common convention in our codebase.

```suggestion
  global DUHDocument:         Path
```

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
...
+  global OMFile:              Unit => Path
```
richardxia: Is there a reason that this a `Unit => Path` instead of just a `Path`?

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
...
+  global OverwriteExisting:   Boolean
+  global OnlyBaseHeader:      Boolean
+
+# create a BasicDriverPlan
+# duhFile => path to your duh file (device.json5)
+# OMFile => a function to produce the OMFile, takes Unit as an arg
+# VendorName => the vendor name used for the device
+# DeviceName => the name of the device
+# LocalMetalDirectory => the metal/drivers directory in the device's repository
+global def makeBasicDriverPlan duhFile omFile vendorName deviceName localMetalDirectory dutName=
```
richardxia: Please put a space before the `=`.

```suggestion
global def makeBasicDriverPlan duhFile omFile vendorName deviceName localMetalDirectory dutName =
```

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
...
+  BasicDriverPlan duhFile omFile vendorName deviceName localMetalDirectory dutName True False
+
+# This function provides an external interface through the
+# publish/subscribe varible 'basicDriverPlans'
```
richardxia:
```suggestion
# publish/subscribe variable 'basicDriverPlans'
```

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
...
+# If you want automatic basic driver creation, create
+# a BasicDriverPlan using makeBasicDriverPlan
+# Then build the drivers by calling
+# crateBasicDriver
```
richardxia:
```suggestion
# createBasicDriver
```

Also, the current locations of line breaks in the comments seem a little bit awkward and arbitrary here. Can we format these so that they read more naturally?

program/freedom-metal-program.wake
```diff
@@ -179,14 +179,31 @@ global def freedomMetalDUTProgramCompiler =
   def score        = (\_ \_ 1.0) # TODO
 
   def imp dut programCompileOptions =
+    def omFile =
+      dut.getDUTObjectModel
+      | prettyJSON
+      | write "build/api-generator-sifive/{dut.getDUTName}/programs/omfile.json"
+    def driverFiles =  match (createBasicDriverWithOMFile dut.getDUTName omFile)
+      Some a = a
+      None   = makeNilDriverOutput
```
richardxia: Is this `makeNilDriverOutput` function really necessary? I feel like either `createBasicDriverWithOMFile` should always return a Nil Driver, eliminating the need to declare a global `makeNilDriverOutput` function, or we should force all callers of `createBasicDriverWithOMFile` to deal with the `None` result. My preference is the latter, since `None` conveys something different (no driver published) from `DriverOutput Nil Nil` (driver published but created no files for some reason).

program/freedom-metal-program.wake
```diff
@@ -179,14 +179,31 @@ global def freedomMetalDUTProgramCompiler =
...
     def outputDir = programCompileOptions.getProgramCompileOptionsOutputDir
     def installDir = "build/freedom-metal/{dut.getDUTName}"
     def metalInstall =
       fixupDTS dut.getRocketChipDUTDTS None
       | (generateMetalFromDts _ dut.getDUTName installDir)
     def gccProgramPlan =
       def outputFile = "{outputDir}/{programCompileOptions.getProgramCompileOptionsName}.elf"
-      programCompileOptionsToGCCProgramPlan outputFile programCompileOptions
+      def newCFiles =
+        driverFiles.getDriverOutputCFiles ++
```
richardxia: If you do the suggestion I made above with not having `makeNilDriverOutput`, and if `driverFiles` is an `Option DriverOutput`, then this could be written as `driverFiles | omap getDriverOutputCFiles | getOrElse Nil`.

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+    """
+    :return:
+    """
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "-o",
+        "--object-model",
+        type=argparse.FileType('r'),
+        help="",
```
richardxia: Please don't fill in dummy values for the `help` string. If you're not going to put one in, then just omit the argument.

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
...
+      then "-x", Nil
+      else Nil
+    def onlyBaseHeader =
+      if basicDriverPlan.getBasicDriverPlanOnlyBaseHeader
+      then "-b", Nil
+      else Nil
+    overwrite ++ onlyBaseHeader
+  def cmd =
+    "{here}/../scripts/generate_drivers.py".simplify,
+    "-o",
```
richardxia: Please use the long form of command line options in scripts, since the short form is cryptic and using the long form acts as a form of documentation. Code is written once but read many times, so please favor code that easier to read rather than easier to write.

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+
+    parser.add_argument(
+        "-d",
+        "--duh-file",
+        type=argparse.FileType('r'),
+        help="",
+    )
+
+    parser.add_argument(
+        "-v",
```
richardxia: Please don't use `-v` for `--vendor`. `-v` is most commonly used in command line programs for `--version`, and sometimes `--verbose`.

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
+#!/home/stillson/lpy/bin/python
```
richardxia: Please remove this.

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
...
+    "-d",
+    basicDriverPlan.getBasicDriverPlanDuhFile.getPathName,
+    "-v",
+    basicDriverPlan.getBasicDriverPlanVendorName,
+    "-D",
+    basicDriverPlan.getBasicDriverPlanDeviceName,
+    "-m",
+    basicDriverPlan.getBasicDriverPlanLocalMetalDirectory,
+    Nil
+    ++ extraFlags
```
richardxia: You don't need the `Nil` and `++` here; you can simply put the `extraFlags` after the comma from `basicDriverPlan.getBasicDriverPlanLocalMetalDirectory,` since `,` is the prepend/cons operator in Wake.

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
...
+  def includeDirs =
+    "{metalDir}",
+    Nil
+  DriverOutput
+  cFiles
+  includeDirs
+
+def makeBasicDrivers basicDriverPlan =
+  basicPlanToPlan basicDriverPlan
+  | addPythonEnv ("{here}/../scripts/generate_drivers_env".simplify)
```
richardxia: I think it'd be better to include this `addPythonEnv` line to `basicPlanToPlan`, since that would allow that function for fully describing the `Plan`, and then the rest of this `makeBasicDrivers` is about running and handling the outputs of the Job.

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
...
+
+tuple DriverOutput =
+  global CFiles:      List Path
+  global IncludeDirs: List String
+
+global def makeNilDriverOutput =
+  DriverOutput Nil Nil
+
+def makeDriverOutput basicDriverPlan jobOutputs =
+  def ifPathCfile path =
```
richardxia: Please capitalize the "F" in File, since proper camelCase would have that capitalized, just like how you have a `cFiles` variable lower down.

```suggestion
  def ifPathCFile path =
```

dut/drivers.wake
```diff
@@ -0,0 +1,109 @@
...
+    path.getPathName
+    | matches `.*\.[c,h]`
+  def cFiles = filter ifPathCfile jobOutputs
+  def metalDir = basicDriverPlan.getBasicDriverPlanLocalMetalDirectory
+  def deviceName = basicDriverPlan.getBasicDriverPlanDeviceName
+  def includeDirs =
+    "{metalDir}",
```
richardxia: This interpolated string seems unnecessary, since there's nothing in the interpolated string but a single string itself.
```suggestion
    metalDir,
```

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+    args = handle_args()
+
+    object_model = json.load(args.object_model)
+    duh_info = json5.load(args.duh_file)
+    vendor = args.vendor
+    device = args.device
+    metal_dir = args.metal_dir
+    check_existence = args.check_existence
+    base_header_only = args.base_header_only
+    m_dir_path = Path(metal_dir)
```
richardxia: You should just set the `type=` argument to [`ArgumentParser.add_argument()`](https://docs.python.org/3/library/argparse.html#the-add-argument-method) to `Path` if you want this to be a Path object.

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+
+
+def main():
+    """
+
+    :return: exits 0 on success, 1 on failure
+    """
+
+    # debugging
+    print(sys.argv, file=sys.stderr)
```
richardxia: Please remove debug printing.

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+    metal_dir = args.metal_dir
+    check_existence = args.check_existence
+    base_header_only = args.base_header_only
+    m_dir_path = Path(metal_dir)
+
+    # process register info from duh
+    addr_blocks = (i['addressBlocks'][0] for i in
+                   find_component(duh_info, 'addressBlocks'))
+
+    regs = next(addr_blocks)['registers']
```
richardxia: I don't think it's correct to only look at the first address block's registers. There are several issues with this:

- A device could have multiple address blocks, and the first one may not necessarily be the one that contains registers, since it could have a memory-like region.
- A device could have multiple address blocks, each with its own set of unique registers.

Also, I think using a generator comprehension and calling `next()` on it to get the first result is an unnecessary optimization, and it's going to be less readable to people who are less familiar with Python.

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+
+        if not base_header_only:
+            header_file_path = m_hdr_path / f'{vendor}_{device}{index}.h'
+            if not base_header_only and \
+                    not header_file_path.exists() or not check_existence:
+                header_file_path.write_text(
+                    generate_metal_dev_hdr(vendor, device, index, base, reglist))
+            else:
+                print(f"{str(header_file_path)} exists, not creating.",
+                      file=sys.stderr)
```
richardxia: I think this `main()` function is too long and also way too deeply nested. In a followup PR, can you break this up into separate functions so that it's easier to follow?

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+
+    reglist: t.List[Register] = []
+    for a_reg in regs:
+        name: str = a_reg['name']
+        offset: int = a_reg['addressOffset'] // 8
+        width: int = a_reg['size']
+        if width not in (8, 16, 32, 64):
+            raise Exception(f'Invalid Register width {width}, register: '
+                            f'{name}, not 8, 16, 32, or 64'
+                            f'fix register width in duh file')
```
richardxia: This message is going to look like:

```
Invalid Register width 7, register: foo, not 8, 16, 32, or 64fix register width in duh file
```

Aside from the missing space before the word "fix", this just doesn't read clearly. The register's name is interposed between the invalid width value and the valid width values. There is a colon that is the strongest English punctuation mark and that partitions the message into two parts, but the parts are not logically grouped, since the register's name, the valid register widths, and the instruction on how to fix the issue are all grouped tighter than the label "register" and the register's name. There is no punctuation that separates the instruction on fixing the problem from the description of the problem itself. The word "Register" is arbitrarily capitalized in the middle of a sentence in its first occurrence.

Readability is important, and the details matter especially in cases where there are user-facing messages. Please write this error message in a way that is more readable. My suggestion is:

```python
raise Exception(
    f'Register {name} has invalid register width {width}; Expected 8, 16, 32, or 64. '
    'Please correct this in the DUH document.'
)
```

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+
+        reglist.append(Register(name, offset, width))
+
+    # parse OM to find base address of all devices
+    memory_regions = find_json_field_name(object_model, '_types', 'OMMemoryRegion')
+    devices = find_json_field_name(memory_regions, 'name', f'{device}.*')
```
richardxia: You can't make this assumption about the memory region name. The correct way of querying for this is to find the Object Model components which match the device using the `_types` field, and then to locate the `OMMemoryRegions` associated with it. The `OMMemoryRegions` are statically known to be in a member field named [`memoryRegions`](https://github.com/chipsalliance/rocket-chip/blob/e627d93fa64c6ea1640915661316b04a91f86029/src/main/scala/diplomaticobjectmodel/model/OMDevice.scala#L6).

Also, naming this variable `devices` is misleading because in the Object Model the "device" is the parent object containing the `OMMemoryRegion`s.

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+
+    parser.add_argument(
+        "-m",
+        "--metal-dir",
+        help="",
+    )
+
+    parser.add_argument(
+        "-x",
+        "--check-existence",
```
richardxia: This isn't a clear argument name. What is this checking the "existence" of? And what is the consequence of checking the existence of it?

I think a better name would be `--overwrite-existing`, which makes it clearer that 1) this is about the existence of the drivers, since "overwrite" ties this flag to the primary action of the script itself and 2) the behavior is that if the drivers already exist, then if this flag is true, the drivers will be overwritten.

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+        name = a_reg.name.lower()
+        cap_name = a_reg.name.upper()
+        size = a_reg.width
+
+        write_func = f"""
+            void {device}_{name}_write(uint32_t *{device}_base, uint{size}_t data)
+            {{
+                volatile uint32_t *control_base = {device}_base;
+                METAL_{cap_device}_REGW(METAL_{cap_device}_{cap_name}) = data;
+            }}"""
```
richardxia: Be careful about newlines. I think you should put the closing triple quotes on their own line here, since that'll ensure that there is a trailing newline that facilitates concatenating the string with other strings, and it will at least be consistent with the `read_func` defined below.

scripts/genererate_drivers.py
```diff
@@ -0,0 +1,633 @@
...
+
+    # parse OM to find base address of all devices
+    memory_regions = find_json_field_name(object_model, '_types', 'OMMemoryRegion')
+    devices = find_json_field_name(memory_regions, 'name', f'{device}.*')
+    devices_enumerated = list(enumerate(devices))
+
+    m_hdr_path = m_dir_path / device
+
+    if not m_hdr_path.exists():
+        m_hdr_path.mkdir()
```
richardxia: You can unconditionally call [`.mkdir(exist_ok=True)`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir) to avoid having to put the guard as a separate if statement.

wit-manifest.json
```diff
@@ -13,5 +13,10 @@
         "commit": "50de8a34c19c12de5066cd7ada50ebb5f5b2ea26",
         "name": "rocket-chip",
         "source": "git@github.com:chipsalliance/rocket-chip.git"
+    },
+    {
+        "commit": "deadbeefdeadbeefdeadbeefdeadbeefdeadbeef",
+        "name": "wake-languages-sifive",
+        "source": "git@github.com:sifive/wake-languages-sifive.git"
```
richardxia: We should update this to point at the renamed version of this, which is `wake-languages-sifive.git`. Also, we should make sure to point it at a real commit.

dut/drivers.wake
sistillson: If you get the OMFile from building the RTL, you need to delay execution.

dut/drivers.wake
sistillson: fixed

program/freedom-metal-program.wake
sistillson: I chose the former, because it's easier to deal with, and any issues with driver creation will fail earlier.

fixec

program/freedom-metal-program.wake
sistillson: works

scripts/genererate_drivers.py
sistillson: sorry, forgot the help strings.

fixed, and for a special bonus add required to the appropriate args

dut/drivers.wake
sistillson: fixed

scripts/genererate_drivers.py
sistillson: fixed

scripts/genererate_drivers.py
sistillson: fixed

dut/drivers.wake
sistillson: fixed

dut/drivers.wake
sistillson: fixed

dut/drivers.wake
sistillson: fixed

dut/drivers.wake
sistillson: fixed

scripts/genererate_drivers.py
sistillson: fixed

scripts/genererate_drivers.py
sistillson: pr created

scripts/genererate_drivers.py
sistillson: fixed

scripts/genererate_drivers.py
sistillson: fixed

scripts/genererate_drivers.py
sistillson: fixed

wit-manifest.json
sistillson: fixed

scripts/genererate_drivers.py
sistillson: find_json_field_name will search again all elements in a list if they are a value a dictionary.

also, that is essentially what I'm doing because devices searches over memory regions

scripts/genererate_drivers.py
sistillson: fixed

scripts/genererate_drivers.py
sistillson: cleaned up, but I would need an example of what you are talking about to make sure it worked

dut/drivers.wake
```diff
@@ -0,0 +1,107 @@
...
+  makePlan cmd visibleFiles
+  | addPythonEnv ("{here}/../scripts/generate_drivers_env".simplify)
+
+tuple DriverOutput =
+  global CFiles:      List Path
+  global IncludeDirs: List String
+
+def foo x =
+  def bar = format x | println
+  x
```
richardxia: Please remove debugging code.

dut/drivers.wake
```diff
@@ -0,0 +1,107 @@
...
+
+def makeDriverOutput basicDriverPlan jobOutputs =
+  def ifPathCFile path =
+    path.getPathName
+    | matches `.*\.[c,h]`
+  def cFiles = filter ifPathCFile jobOutputs
+  def metalDir = basicDriverPlan.getBasicDriverPlanLocalMetalDirectory
+  def deviceName = basicDriverPlan.getBasicDriverPlanDeviceName
```
richardxia: This line seems to be unused now.

dut/drivers.wake
richardxia: I'm still confused about why we need to delay the execution here. Couldn't you just delay the construction of the BasicDriverPlan until after OM is constructed?

dut/drivers.wake
```diff
@@ -0,0 +1,107 @@
+
+tuple BasicDriverPlan =
+  global DUHDocument:         Path
+  global OMFile:              Unit => Path
+  global VendorName:          String
+  global DeviceName:          String
+  global LocalMetalDirectory: String
+  global DUTName:             String
```
richardxia: I'm a bit confused by this `DUTName` argument. Why is this necessary, and why are we doing string matching on it? Why can't we automatically use the same driver for any DUT as long as the driver, not the DUT, matches?

I think I'm starting to realize I don't really understand at a high level how this mechanism is intended to work. Can you explain the relationships between a hardware IP block (also known as a "device"), a driver, a DUT, and a BasicDriverPlan? In particular, it seems like right now a BasicDriverPlan is one-to-one with the cross product of IP blocks and DUTs, which doesn't sound like it'll scale. How do you create multiple drivers for one DUT? How do you create a driver for one block that can be reused in multiple DUTs?

dut/drivers.wake
```diff
@@ -0,0 +1,107 @@
...
+
+def makeBasicDrivers basicDriverPlan =
+  basicPlanToPlan basicDriverPlan
+  | runJob
+  | getJobOutputs
+  | makeDriverOutput basicDriverPlan
+
+# External wake target to create basic drivers
+# deviceName => the name of the DUT plan
+global def createBasicDriver dutPlanName =
```
richardxia: The comment says the argument name is `deviceName`, but the actual code says the argument is `dutPlanName`. The former makes more sense to me, but it appears that your code is actually making use of the latter. Why is it that we are filtering based on the DUT name rather than the device name?

dut/drivers.wake
```diff
@@ -0,0 +1,107 @@
...
+  getDevicePlan dutPlanName
+  | omap makeBasicDrivers
+
+# wake target to create basic drivers from internal code where
+# omfile is generated
+# deviceName => the name of the DUT Plan
+global def createBasicDriverWithOMFile dutPlanName omfile =
```
richardxia: It's not clear to me when you would use this function vs the `createBasicDriver` function. What is the difference in terms of the output between the two?

richardxia (changes requested):

dut/drivers.wake
sistillson: fixed

dut/drivers.wake
sistillson: removed

dut/drivers.wake
sistillson: They are basically the same. CreateBasicDriverWithOMFile is used when you actually have an omfile. But, it's just a convenience function really to deal with the omfile being delayed.

dut/drivers.wake
sistillson: We need to build the DUT to get the OM to feed into the driver creation. Comment is wrong. 

I think this will change, so I will put off fixing it for a bit.

dut/drivers.wake
albertchen-sifive: also I don't think the tuple that we publish to `basicDriverPlans` should be the same tuple that is used as the plan for `makeBasicDrivers`. because like you said the OMFile isn't available until after the dut is elaborated same with things like the `BSPDirectory` and `DUTName`. There are also options that don't really make sense in the context of your driver like `OverwriteExisting` or `BaseHeader`.

dut/drivers.wake
```diff
@@ -0,0 +1,188 @@
...
+# duhFile => path to your duh file (device.json5)
+# OMFile => a function to produce the OMFile, takes Unit as an arg
+# VendorName => the vendor name used for the device
+# DeviceName => the name of the device
+# LocalMetalDirectory => the metal/drivers directory in the device's repository
+global def makeBasicDriverPlan duhFile omFile vendorName deviceName localMetalDirectory bspDirectory compat dutName =
+  BasicDriverPlan duhFile omFile vendorName deviceName localMetalDirectory bspDirectory compat dutName True True True
+
+# memoize this because trees in wake are slow
+target basicDriverCompatibilityTrees Unit =
```
richardxia: I've never properly learned how `target` works, since there's no documentation on it. Is it safe to declare a function as a `target` when it only accepts a `Unit` and within the body, there's a `subscribe`?

@albertchen-sifive, @jackkoenig, or @terpstra, can you explain how `target` works? Even better is if we could properly document it in `wake.git`.

Also, another I've been wondering is whether the order of items coming back from a `subscribe` is deterministic.

dut/drivers.wake
```diff
@@ -0,0 +1,188 @@
+
```
richardxia, line 1: I think we can split this file into two separate files:

1. A file that only deals with slurping up drivers that have already been created and committed to source control and with generating configuration-specific header files.
2. A file that only deals with automatically generating driver files from the DUH document.

The two steps are unrelated to each other, and there's no need for them to share any code. The latter shouldn't be dependent on a number of other things, including the object model. I want to submit a PR against this branch demonstrating how this could work.

dut/drivers.wake
albertchen-sifive: I think it's fine to have a `subscribe` within a `target` since wake has to "elaborate" all the pub/subs at compile time in order to check for circular dependencies?

albertchen-sifive: target explanation https://github.com/sifive/wake/issues/124

I don't think the final implementation has changed from how it is described in the issue other than that target will panic if your "ignored" arguments differ for the same "key" arguments.

richardxia: Thanks, that's a very useful ticket to learn about `target`!

scripts/generate_drivers.py
```diff
@@ -0,0 +1,600 @@
...
+        type=argparse.FileType('r'),
+        help="The path to the object model file",
+    )
+
+    parser.add_argument(
+        "-d",
+        "--duh-document",
+        type=argparse.FileType('r'),
+        help="The path to the DUH document",
+        required=True
```
albertchen-sifive, line 373: why is this required? I thought you could you this script to just generate headers using just the object model?

albertchen-sifive: same with `--metal-dir`/`--bsp-dir`. It looks to me like these are just output directory args for the basic drivers and base headers respectively. why are they both required when you should be able to run either flow individually?

sistillson: because you might run them at the same time.

albertchen-sifive: but we want to be able to run them separately though. for the PIO block we only have access to the Object Model and we only want to generate the base header, how do we do that if we are also required to provide the DUH document?

PR #5, opened 2019-10-15
lookup drivers based on compatible strings

albertchen-sifive:
replace `DUTName` field of `BasicDriverPlan` with a function that returns `True` if that driver is compatible based on the compatible string lists from the DTS. driver functions return lists of drivers now instead of a single driver.


sistillson (commented):
What if you have two drivers which are compatible for a block, how do you decide which to use?

dut/drivers.wake
```diff
@@ -87,16 +85,20 @@ def makeBasicDrivers basicDriverPlan =
 
 # External wake target to create basic drivers
 # deviceName => the name of the DUT plan
-global def createBasicDriver dutPlanName =
-  getDevicePlan dutPlanName
-  | omap makeBasicDrivers
+global def createBasicDrivers dts =
```
sistillson: dts or compatibleList?

albertchen-sifive:
> What if you have two drivers which are compatible for a block, how do you decide which to use?

hmm didn't think of that. probably `IsCompatible` should be something like `List String => Option Double` or something? so we do the lookup for each compatible list we find an pick the highest score for each one?

dut/drivers.wake
```diff
@@ -87,16 +85,20 @@ def makeBasicDrivers basicDriverPlan =
...
-global def createBasicDriverWithOMFile dutPlanName omfile =
-  def omfileArg _ = omfile
-  getDevicePlan dutPlanName
-  | omap (setBasicDriverPlanOMFile omfileArg)
-  | omap makeBasicDrivers
-  | getOrElse (DriverOutput Nil Nil)
+global def createBasicDriversForDUT dut =
+  def dts = dut.getDUTDTS
+  def omFile = dut.getDUTObjectModelFile
+  def compatibleLists = Nil
```
albertchen-sifive, line 143: need to fill in the DTS parsing stuff here

dut/drivers.wake
```diff
@@ -5,7 +5,7 @@ tuple BasicDriverPlan =
   global VendorName:          String
   global DeviceName:          String
   global LocalMetalDirectory: String
-  global DUTName:             String
+  global IsCompatible:        List (List String) => Boolean
```
sistillson: instead of having every driver write an IsCompatible function, wouldn't it be better to have the driver just supply the list of compatability strings, in order?

dut/drivers.wake
albertchen-sifive: compatibleList I guess? though I don't see why we need this at all since this function isn't used anywhere

dut/drivers.wake
albertchen-sifive: how do we pick which driver to use if there are multiple matches?

sistillson: if there is an exact match, use that. There should only be one. if not go to the next and match on that. The second string is more generic.

albertchen-sifive: so fail if there is more than one match?

sistillson: The spec is vague. 

I would say fail and let it be know why it's failing so that it can be fixed.

albertchen-sifive: ok talked with Richard, and your suggestion sounds good. we'll make this a list of strings and just match in order of the compatible list in the dts. if there are multiple matches just pick one and may print a warning.

dut/drivers.wake
```diff
@@ -5,7 +5,7 @@ tuple BasicDriverPlan =
   global VendorName:          String
   global DeviceName:          String
   global LocalMetalDirectory: String
-  global IsCompatible:        List (List String) => Boolean
+  global IsCompatible:        Tree String
```
sistillson: Why are you using a tree for two strings?

albertchen-sifive:
@sistillson @richardxia updated to allow publishing driver implementations that are automatically included when test programs are compiled

dut/drivers.wake
```diff
@@ -1,11 +1,38 @@
+tuple DriverImplementation =
+  global IsCompatible: List String
+  global Sources:      List Path
```
sistillson, line 3: you also need a list of strings for include directories

PR #6, opened 2019-10-16
Add dependency on scribble-testsocket-sifive

richardxia:
This is a WIP branch which we are using to test all of these new flows together. Note that this does point to the scribble-testsocket-sifive.git repo, which is still private at the moment, but once we get everything in better shape we'll open source everything.


sistillson (approved):
lgtm

PR #9, opened 2019-10-22
merge base header pub/sub with driver pub/sub

albertchen-sifive:
this merges the old `basicDriverPlans` pub/sub thing with the `driverImplementations` pub/sub thingy. also adds some fields to the `DriverImplementation` tuple, replaces `basicPlanToPlan` with `makeBaseHeader` that runs the same python script but only uses the basic header flow, and rearrange some stuff in the python script.


sistillson (commented):
The generate_driver changes are generally going to make it harder for me to get to a final version.

sistillson:
is there any way to generate driver code with wake?

Otherwise it looks good.

albertchen-sifive:
> is there any way to generate driver code with wake?
> 
> Otherwise it looks good.

you mean running the `generate_drivers.py` script? I don't think we want to run that in wake, because it's more of a duh-scala kind of flow where you run the script manually and make some modifications to the generated code then publish your completed driver sources to the `driverImplementations` pub/sub thingy.

sistillson:
Sounds good to me
Approved

PR #10, opened 2019-10-24
Publish driver generator dependencies to preinstall topic.

richardxia:
This allows those files to be fetched as an independent step before running the later build steps.

This specifically addresses a Travis CI check that ensured that most of the build does not rely on the internet: https://github.com/sifive/block-pio-sifive/pull/21

I have to say, I really like how we have that Travis CI check in place, since it allowed us to catch this, and that we have this `preinstall` pub/sub topic, which allows us to very easily front-load certain build steps.

I'm wondering if this is even something we could further generalize for separating out steps that may have other specific requirements that we would want to run upfront, such as a Scala compilation step that allows us to compile all code first and cache it.


sistillson:
Nice, but maybe this can be generalized in api-languages-sifive?

richardxia:
> Nice, but maybe this can be generalized in api-languages-sifive?

Maybe, but I'm wondering if there should be more "lifecycle hooks" beyond just the "preinstall" step. Something like what Travis CI has in terms of specific points in time that you can inject code: https://docs.travis-ci.com/user/job-lifecycle

Anyway, I've tested that [this works in block-pio-sifive](https://travis-ci.com/sifive/block-pio-sifive/builds/133314978) when I set up a temporary branch to point to this PR, so I'm going to merge this in now.

PR #13, opened 2019-11-11
Add resources to sim options

albertchen-sifive:
Add a tuple field to allow passing extra resources to simulation compile and run jobs. Also get rid of a bunch of helper functions that weren't all that useful.

### DIFF
--- a/sim/dut-simulator.wake
+++ b/sim/dut-simulator.wake
@@ -90,9 +90,10 @@ tuple DUTSimCompileOptions =
   global Defines:        List NamedArg
   global SourceFiles:    List Path
   global Plusargs:       List NamedArg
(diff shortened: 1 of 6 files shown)


sim/dut-simulator.wake
```diff
@@ -90,47 +90,53 @@ tuple DUTSimCompileOptions =
   global Defines:        List NamedArg
   global SourceFiles:    List Path
   global Plusargs:       List NamedArg
+  global Resources:      List String
```
richardxia: Might be worth adding a comment here that describes the fact that this should be the same concept of "Resources" as in a Wake `Plan`.

richardxia (approved):
LGTM!

PR #14, opened 2019-11-13
get extra files and includes from environment variables

albertchen-sifive:
update the `verilator_wrapper` script to pull in arguments from `WAKE_IP_RESOURCE_INCLUDES` and `WAKE_IP_RESOURCE_FILES` environment variables. Wake runners can use these environment variables to provide extra verilog files and includes for IP that are not installed in the workspace.

### DIFF
--- a/scripts/verilator_wrapper
+++ b/scripts/verilator_wrapper
@@ -2,4 +2,22 @@
 
 VERILATOR_PATH=$(which verilator)
 unset VERILATOR_ROOT
-perl $VERILATOR_PATH "$@"
+
+OIFS="$IFS"
+IFS=":"
+
+includes="$WAKE_IP_RESOURCE_INCLUDES"
+include_args=( )
+for include in $includes; do
+  include_args+=( +incdir+"$include" )
+done
+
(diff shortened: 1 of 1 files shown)


jackkoenig:
Is it that the normal wake API can't accept input files that aren't in the workspace? Is it better to do this with environment variables and forcing the runner to do it rather than just passing the absolute paths?

richardxia (approved):

albertchen-sifive:
> Is it that the normal wake API can't accept input files that aren't in the workspace? Is it better to do this with environment variables and forcing the runner to do it rather than just passing the absolute paths?

the wake API can accept files not in the workspace but we would rather not deal with absolute paths outside of the runner for better relocatability.

rmac-sifive:
Thanks @albertchen-sifive. For my and the SoC verif team's benefit: mechanically how does this work? Should we set `WAKE_IP_RESOURCE_INCLUDES=/sifive/sip/synopsys/i2c:/sifive/sip/synopsys/i2s`, etc.?

PR #15, opened 2019-11-14
Systemc ipdelivery

EricCheng2222:
add some .wake files for systemc ipdelivery such as systemc.wake, systemc-sim.wake, and a NEED-HACK file to explain places need to be hacked in order to build systemc with wake flow. However, the version I pushed is not working because I took away the hacking part


i-am-ahung:
I am referring to this email........

PR #16, opened 2019-11-14
Extend generate headers to work with oddly sized interrupt fields

sistillson:
add interupt information

Made information more complete.

### DIFF
--- a/scripts/generate_header.py
+++ b/scripts/generate_header.py
@@ -215,10 +215,21 @@ class Interrupt:
         return Interrupt(number, name)
 
 
-def generate_interrupt_list(object_model: JSONType) -> t.List[Interrupt]:
+def generate_interrupt_list(object_model: JSONType, device: str) -> t.List[Interrupt]:
+
+    def type_match(dev: str, types: t.List[str]):
+        d_str = dev.lower()
+        for a_type in types:
+            if a_type.lower().endswith(d_str):
+                return True
+        return False
+
     p = walk(object_model)
     p = filter(lambda x: '_types' in x, p)
-    p = filter(lambda x: f'OMInterrupt' in x['_types'], p)
+    p = filter(lambda x: type_match(device, x['_types']), p)
+    p = walk(list(p))
+    p = filter(lambda x: '_types' in x, p)
+    p = filter(lambda x: 'OMInterrupt' in x['_types'], p)
 
     rv = []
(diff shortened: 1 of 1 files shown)


scripts/generate_header.py
```diff
@@ -39,15 +42,12 @@ class Register:
     name: str
     offset: int  # in bytes
     width: int  # in bits
+    group: str
```
richardxia: What exactly is this `group`, and how is it different from the `name`? In the Chisel generators that we have, what we call a "RegFieldGroup" [[1]](https://github.com/chipsalliance/rocket-chip/blob/bb6247aeb82cd1f0730a3249cf1a283b0b94c342/src/main/scala/regmapper/RegField.scala#L41) [[2]](https://github.com/chipsalliance/rocket-chip/blob/f2cb2bba25406de7f48229741883b2bc9181dd92/src/main/scala/devices/debug/Debug.scala#L478) is basically the same thing as a "register" in most cases.

Is what you are calling a Register here closer to what we are calling a RegisterField in Chisel? If that's the case, can we at least write a comment somewhere in the code that describes the differences in terminology? I think it's a significant enough difference that it is worth explicitly commenting because the term "Register" is being used in two different ways.

scripts/generate_header.py
```diff
@@ -169,43 +177,97 @@ def generate_offsets(device_name: str, reg_list: t.List[Register]) -> str:
...
+
+    for an_interrupt in int_list:
+        number = an_interrupt.number
+        if an_interrupt.name:
+            name = an_interrupt.name.upper().replace(' ', '')
+            rv.append(f'#define {dev}_{name}_IT {number}')
+        interrupts.append(an_interrupt.number)
+
+    rv.append(f'#define {dev}_INTERRUPT_COUNT {len(interrupts)}')
+    rv.append(f'#define {dev}_INTERRUPTS {{ {",".join(map(str,interrupts)) } }}')
```
richardxia: I'm not sure if this will work well if there are multiple instances of a device in a design. Although it might be safe in most cases to assume that the relative offsets of the interrupt numbers are consistent across different device instances, they will definitely have different absolute numbers.

We probably need a data structure here that describes at least the base of the block of interrupt numbers that is assigned to the device.

richardxia (changes requested):

richardxia (approved):
LGTM!

PR #18, opened 2019-11-19
check for bad paths for better error reporting

albertchen-sifive:
`runSim` tends to fail with unhelpful error messages because we aren't checking for bad paths

### DIFF
--- a/sim/simulator.wake
+++ b/sim/simulator.wake
@@ -1,7 +1,3 @@
-def mustPass x = match x
-  Pass a = a
-  Fail b = panic b.format
-
 tuple TestFailed =
   global Msg:    String
   global Status: Option Integer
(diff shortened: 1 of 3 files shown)


richardxia (approved):
This LGTM, but I feel like the fact that we have to make all of these kinds of changes implies that something is a bit off about the current design of Jobs and Paths/BadPaths in Wake.

albertchen-sifive:
> This LGTM, but I feel like the fact that we have to make all of these kinds of changes implies that something is a bit off about the current design of Jobs and Paths/BadPaths in Wake.

yes. some helpers at least would be nice. something like `getJobPath "my/file" myJob"` that would do the `isJobOk` check and find a matching path.

PR #19, opened 2019-11-19
add option to specify program entry, rom, and ram

albertchen-sifive:
Updates `freedomMetalProgramCompiler` to add `metal,rom` and `metal,ram` chosen nodes.
Adds `Entry`, `Rom`, and `Ram` fields to `TestProgramPlan` the tuple.

### DIFF
--- a/program/freedom-metal-program.wake
+++ b/program/freedom-metal-program.wake
@@ -1,9 +1,10 @@
-# FIXME
-def addMetalMemory type memSubnode inDTSContents =
+# takes a device tree property name and a subnode type and injects a reference
+# to the first occurance of that subnode type to the chosen node
+def addChosenRef name socSubNode inDTSContents =
   def subnodeOpt =
-    match "({memSubnode}@[0-9]+)".stringToRegExp
+    match "({socSubNode}@[a-fA-F0-9]+)".stringToRegExp
       Pass regex =
-        def nodeRegex = regExpCat (`^\t\t`, `(?:L[0-9]+:\s)?`, regex, `\s{$`, Nil)
+        def nodeRegex = regExpCat (`^\t\t`, `(?:L[a-zA-Z0-9_]+:\s)?`, regex, `\s{$`, Nil)
         inDTSContents
         | map nodeRegex.extract
         | flatten
@@ -13,14 +14,12 @@ def addMetalMemory type memSubnode inDTSContents =
 
   match subnodeOpt
     Some subnode =
-      def asdf =
-        inDTSContents
-        | mapFlat (\a (
-          if matches `^\tchosen\s{$` a
-          then ("\tchosen \{", "\t\tmetal,{type} = <&\{{subnode}\}>;", Nil)
-          else a, Nil
-        ))
-      asdf
+      inDTSContents
+      | mapFlat (\a (
+        if matches `^\tchosen\s{$` a
+        then ("\tchosen \{", "\t\t{name} = <&\{{subnode}\}>;", Nil)
+        else a, Nil
+      ))
     None = inDTSContents
 
 def addEntry (Pair memSubnode offsetInt) inDTSContents =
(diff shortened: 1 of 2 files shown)


program/freedom-metal-program.wake
```diff
@@ -111,7 +135,8 @@ global target fixupDTS dts metalEntryPairOpt =
       | addChosenNode
       | addStdoutChosenFromAlias
       | (omap addEntry metalEntryPairOpt | getOrElse (_))
-      | addHfclk clkFrequency
+      | (omap "rom".addMetalMemory romNode | getOrElse (_))
```
sistillson: did you mean to take out the `addHfclk` call?

sistillson (commented):
seems ok.  I think we need to start putting comments into our wake code...

program/freedom-metal-program.wake
albertchen-sifive: good catch. didn't mean to do that in this PR

program/freedom-metal-program.wake
```diff
@@ -1,4 +1,28 @@
 # FIXME
+def addMetalMemory type memSubnode inDTSContents =
+  def subnodeOpt =
+    match "({memSubnode}@[0-9]+)".stringToRegExp
```
richardxia: I'm assuming that this is supposed to match one of the DTS nodes names that look like `clint@2000000`. If so, then the address on the right side is a hex number and you need to include `a-z` in that pattern.

program/freedom-metal-program.wake
```diff
@@ -1,4 +1,28 @@
...
+      Pass regex =
+        def nodeRegex = regExpCat (`^\t\t`, `(?:L[0-9]+:\s)?`, regex, `\s{$`, Nil)
```
richardxia: I just want to point out that this is making the assumption that all DTS labels are of the form `L[0-9]+`, which is only coincidentally true for the DTS files generated by rocket-chip. If we/they/someone ever decides to change this, it'll break this. You might want to write a more generic regex for matching any kind of label. According to the [spec](https://github.com/devicetree-org/devicetree-specification/releases/download/v0.2/devicetree-specification-v0.2.pdf), the valid characters in a label are `[a-zA-Z0-9_]`.

program/freedom-metal-program.wake
```diff
@@ -1,4 +1,28 @@
...
+        inDTSContents
+        | map nodeRegex.extract
+        | flatten
+        | head
+        | omap ("/soc/{_}")
+      Fail _ = None
+
+  match subnodeOpt
+    Some subnode =
+      def asdf =
```
richardxia: Leftover debugging variable name?

program/freedom-metal-program.wake
```diff
@@ -1,4 +1,28 @@
...
+        inDTSContents
+        | mapFlat (\a (
+          if matches `^\tchosen\s{$` a
+          then ("\tchosen \{", "\t\tmetal,{type} = <&\{{subnode}\}>;", Nil)
```
richardxia: Perhaps you could consider changing the `type` variable to include the `metal,` prefix, so that this is an entire Device Tree property name? This would make it clearer how the specific type name is combined with the `metal,` prefix at the call site.

program/freedom-metal-program.wake
```diff
@@ -1,4 +1,28 @@
 # FIXME
+def addMetalMemory type memSubnode inDTSContents =
```
richardxia: Since this function is doing something reasonably complex, could you add a comment describing at a high level what it's doing?

My best guess is that given a device tree property type name and a node type, it finds the label of the matching node, and then it injects a reference to that label to the `chosen` node.

richardxia (approved):
A few suggestions about making the code a bit more general and about documenting it a bit more, but otherwise this LGTM.

PR #20, opened 2019-11-20
make makeVCSDUTSimulatorPlan global, remove addHfclk

albertchen-sifive:
`addHfclk` is no longer needed since rocket chip has https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/diplomacy/FixedClockResource.scala. I don't think it ever was needed, we just didn't know how to do it in scala. also make `make{VCS|Verilator|Xcelium}DUTSimulatorPlan` global so other people can actually make their own `DUTSimulator`s


richardxia (approved):
LGTM!

PR #21, opened 2019-11-22
Parse memoryRegion from OM

rmac-sifive:


program/program.wake
```diff
@@ -243,7 +243,8 @@ global def makePlusargProgramLoader plusarg memoryRegion score compiler =
     def outputDir = loaderOptions.getProgramLoaderOptionsOutputDir
     def programOptions = loaderOptions.getProgramLoaderOptionsProgramOptions
     def hex =
-      def bitWidth = 64 # TODO: get from object model
+      def bitWidthOpt = getMemoryRegion dut.getDUTObjectModel memoryRegion
+      def bitWidth = getOrElse 32 bitWidthOpt
```
richardxia, line 247: Hey @rmac-sifive, this change is causing block-pio-sifive.git to fail when trying to bump to the latest api-generator-sifive.git because it changed the default value of the hex file bit width from 64 to 32.

Is there a reason it was changed from 64 to 32, or can I change it back? I won't be able to rely on the getting the proper width from the Object Model because in the Test Socket we're not currently emitting an OMSRAM for the mbus memory that the Test Socket uses for loading programs.

cc @sistillson and @albertchen-sifive, since this explains one of the cascade of errors in https://github.com/sifive/block-pio-sifive/pull/48 when attempting to bump api-generator-sifive.

rmac-sifive: Changing this default back to 64 will break some of our internal designs, because they rely on the default being 32. This change made the poor assumption that all configs other than those would have this OM entry. I think the right solution is to remove the default and make the entry required for all configs, but I'd like to hear what @albertchen-sifive thinks

PR #22, opened 2019-12-04
make stdout-path chosen node configurable

albertchen-sifive:
`stdout-path` is set to the first `serial` node found in the DTS. however the `SimUART` and the real `UART` from sifive-blocks have the same DTS name, which can mess up the `stdout-path`.

This PR adds a `StdoutPath` option to the `TestProgramPlan` tuple to enable configuring this chosen node like we do for entry, ram, and rom. It also changes the default stdout-path node to `simSerial` instead of `serial` in `makeTestProgramPlan` since we usually want `stdout-path` to be the `SimUART` for test programs.

This will break old test plans that expect stdout-path to look for a `serial` node.

### DIFF
--- a/program/program.wake
+++ b/program/program.wake
@@ -97,9 +97,9 @@ tuple ProgramCompileOptions =
   global OutputDir:   String
   global Filter:      DUTProgramCompiler => Boolean
   global Entry:       Option (Pair String Integer)
-  global Rom:         Option String
-  global Ram:         Option String
-  global StdoutPath:  Option String
+  global Rom:         Option String # name of the DTS node to set the 'metal,rom' chosen node to
+  global Ram:         Option String # name of the DTS node to set the 'metal,ram' chosen node to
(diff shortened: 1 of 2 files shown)


program/program.wake
```diff
@@ -39,6 +39,7 @@ tuple TestProgramPlan =
   global Entry:       Option (Pair String Integer)
   global Rom:         Option String
   global Ram:         Option String
+  global StdoutPath:  Option String
```
richardxia: Could we give this a slightly different name? If I'm reading the code correctly, it looks like this is meant to be the node name in DTS. However, when I just read the name of this property, this sounds like it would be the filepath to which you would output the stdout of the process. In either case, let's at the very least document what this means.

albertchen-sifive:
oops sorry I was a little over-eager and merged

richardxia:
> oops sorry I was a little over-eager and merged

Haha, it's all right, it looks functionally correct to me.

PR #23, opened 2019-12-11
Change default hex file width back to 64.

richardxia:
I left a comment in https://github.com/sifive/api-generator-sifive/pull/21 just now, but if we can change this default value back to 64, this will allow https://github.com/sifive/block-pio-sifive/pull/48 to pass.


albertchen-sifive (approved):

albertchen-sifive:
actually, maybe we should figure out why we aren't parsing the correct width in the first place?

richardxia:
We aren't passing the correct width because the current mechanism for passing a width is to scan the Object Model for an OMSRAM object where the `description` matches the `SimProgramLoaderPlan MemoryRegion`. We have two issues:

1. The TestSocket does not construct an OMSRAM for the test harness memory (the thing currently attached to the mbus), and in general does not have anything to represent OM things higher up than the RocketSubsystem.
2. We really don't have a consistent methodology for assigning names to that description field of the `OMSRAM` in the first place, so even if we fixed 1), I'd be hesitant to lock into that without considering how stable that API can be in the first place.

I'm more OK with doing both of these just for the TestSocket, since it's a design where we do manually instantiate and connect a single, unconfigurable memory to the mbus, but I'm not sure how well it would scale to other designs.

Let me know what you think and whether you would prefer that we do 1) and/or 2) instead of just merging in my current PR.

richardxia:
To summarize an offline discussion, we're going to do both 1) and 2) I listed above, and I will just close this PR.

PR #24, opened 2019-12-11
Actually exit with nonzero status when simulation tests fail

richardxia:
Currently if a simulation test fails, we print out a nice string with the results, but the process exits with status code 0, which causes automation like CI to not understand the failure. This PR does a bit of refactoring so that it now returns a `Result` object with a `Pass`/`Fail` corresponding to whether any tests actually failed.

You'll want to review this PR with whitespace diffs ignored, since that makes it look scarier than it actually is.


top.wake
```diff
@@ -47,19 +47,48 @@ global def bitstreamTarget board dut buildRoot =
...
+  def simOutputs = map (singleSimOutput simName waves dut buildRoot) dutTests
+
+  def simCheckerOptions programName =
+    def filter = (\_ True)
+    def outputDir = buildRoot.getBuildTreeReportDir simName programName
+    makeSimCheckerOptions filter outputDir
+
+  def allPassed =
+    simOutputs
+    | zip programNames
```
sistillson: my only worry is using zip here. if simOutputs and programNames get out of sync in the future, this could be a problem.

And maybe add some comments

sistillson (commented):
Looks ok to me

top.wake
```diff
@@ -47,19 +47,48 @@ global def bitstreamTarget board dut buildRoot =
 
 global def simTarget dut buildRoot simName waves =
   def dutTests = subscribe dutTests | filter (_.getDUTTestFilter dut)
-  map (singleSimTarget simName waves dut buildRoot) dutTests
-  | zip dutTests
-  | map getDUTTestName.editPairFirst
-  | aggregateTestResults
-  | println
+  def programNames = map (_.getDUTTestProgram.getTestProgramPlanName) dutTests
+  def simOutputs = map (singleSimOutput simName waves dut buildRoot) dutTests
```
albertchen-sifive: maybe consider mapping a function that returns the name -> output pair. since these two lists are get zipped together twice.

top.wake
```diff
@@ -47,19 +47,48 @@ global def bitstreamTarget board dut buildRoot =
...
+    def outputDir = buildRoot.getBuildTreeReportDir simName programName
+    makeSimCheckerOptions filter outputDir
+
+  def allPassed =
+    simOutputs
+    | zip programNames
+    | map (match _
+      Pair programName simOutput =
+        simCheckerGetResult dut simChecker simOutput (simCheckerOptions programName)
+    )
```
albertchen-sifive: you can pattern match the lambda argument like `(\(Pair programName simOutput) ...)`

top.wake
```diff
@@ -47,19 +47,48 @@ global def bitstreamTarget board dut buildRoot =
...
+    | forall isTestResultPass
+
+  def testReport =
+    simOutputs
+    | zip programNames
+    | map (match _
```
albertchen-sifive: also here

albertchen-sifive (approved):

richardxia:
Thanks for the feedback and Wake tips. I just force-pushed a change with the updates, as well as a few additional comments.

PR #28, opened 2019-12-13
add from env var to ffile for fpga

erikdanie:
Append on contents of WAKE_IP_RESOURCE_FILES to the generated verilog.F file for fpga builds

### DIFF
--- a/fpga/mcs.wake
+++ b/fpga/mcs.wake
@@ -29,9 +29,15 @@ global target makeBitstream plan =
     def runDir = outputDir
 
     def ffile =
-      map (relative runDir _.getPathName) vsrcs
-      | catWith "\n"
-      | write "{runDir}/verilog.F"
+      def vnames =
+        vsrcs
+        | map (relative runDir _.getPathName)
+        | catWith "\n"
+      def enames =
+        getenv "WAKE_IP_RESOURCE_FILES"
+        | getOrElse ""
+        | replace `:` "\n"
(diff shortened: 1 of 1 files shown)


fpga/mcs.wake
```diff
@@ -29,9 +29,15 @@ global target makeBitstream plan =
...
-      | catWith "\n"
-      | write "{runDir}/verilog.F"
+      def vnames =
+        vsrcs
+        | map (relative runDir _.getPathName)
+        | catWith "\n"
+      def enames =
+        getenv "WAKE_IP_RESOURCE_FILES"
+        | getOrElse ""
+        | replace `:` "\n"
```
albertchen-sifive, line 39: wouldn't this always give `""` unless you have `WAKE_IP_RESOURCE_FILES` set in your actual environment? `getenv` grabs the value of `WAKE_IP_RESOURCE_FILES` from the user environment, not the environment that the job is actually running in.

erikdanie: Oh fuck, that was the reason it is done by script?

I thought wake couldnt read env vars at all, and Wes told me it could so I went ahead with this...

richardxia: Since this won't work as initially intended, should we close out this PR?

erikdanie: yeah

erikdanie:
closing this in favor of https://github.com/sifive/api-generator-sifive/pull/30

PR #30, opened 2019-12-19
add environemnt files and mig ip

erikdanie:
Add mig hook and call vivado with new tcl environment variable

### DIFF
--- a/fpga/mcs.wake
+++ b/fpga/mcs.wake
@@ -13,6 +13,7 @@ global def makeBitstreamPlan name vendor board topModule outputDir =
   def resources = "xilinx/vivado/2018.2", Nil
   BitstreamPlan name vendor board topModule Nil outputDir Nil resources Nil
 
+# TODO - should this be moved to fpga-shells?
 def vc707migHook =
   def name = "vc707_1gb_mig"
   def addResources = "vivado/vc7071gmig/1.0"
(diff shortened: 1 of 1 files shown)


fpga/mcs.wake
```diff
@@ -13,6 +13,11 @@ global def makeBitstreamPlan name vendor board topModule outputDir =
   def resources = "xilinx/vivado/2018.2", Nil
   BitstreamPlan name vendor board topModule Nil outputDir Nil resources Nil
 
+def vc707migHook =
+  def name = "vc707_1gb_mig"
+  def addResources = "vivado/vc7071gmig/1.0"
+  makeBlackBoxHook name addResources
```
albertchen-sifive, line 20: not sure if api-generator-sifive is the best place to put this. maybe this should go in fpga-shells? also I just realized that we should probably be deduping the verilog sources in the vsrcs.f file and in the environment variable. in case these things get published multiple times

albertchen-sifive (approved):

fpga/mcs.wake
erikdanie: > not sure if api-generator-sifive is the best place to put this. maybe this should go in fpga-shells?

I guess it could, but to me, this api-generato-sifive/fpga is essentially the same as fpga-shells for wake files. Why would it matter which one it is in?

> I just realized that we should probably be deduping the verilog sources in the vsrcs.f file and in the environment variable. in case these things get published multiple times

Do you mean doing this in tcl?

erikdanie:
Per discussion with Albert, I will merge this now, and possibly think about moving to fpga-shells in the future. For the de-duping, from a quick test, it seems that vivado does not care if you duplicate a file in the .f that you give it, so I don't think this will be a problem

PR #31, opened 2019-12-23
Update driver generation to support register fields

richardxia:
This makes some changes to the driver generation scripts to support fields within registers. This is a _backwards-incompatible_ change, since the names of the functions it generates are now different.

To summarize the differences in API, basically all functions that were named something like `{device}_{register}` (e.g. `pio_odata`) have been changed to `{device}_{register}_{field}` (e.g. `pio_odata_data`).

The other main change is that now the actual functions that perform the load/stores do some bit twiddling to mask and shift the values you care about into the right-most position. Note that for writes, in order to preserve the values of other fields, we do a load first, and then clear and set the bits for field that we actually care about. This is _not_ thread-safe nor is it atomic; this is because I didn't want to risk this code not working for any architecture target that doesn't support atomics, and if one wanted to, they can hand-modify the output of script.

These two changes were made in https://github.com/sifive/api-generator-sifive/pull/31/commits/2da8a6c630eda8379251f85b8fc40105a77afc73.

The other two minor changes I made were:

- https://github.com/sifive/api-generator-sifive/commit/d60e066c7b4f5c22d21aba946b82c430f4bdfd69 - Add the `jsonref` Python library to the Pipfile so that we can easily dereference [JSON references](https://tools.ietf.org/id/draft-pbryan-zyp-json-ref-03.html).
- https://github.com/sifive/api-generator-sifive/commit/2c28c60fe06db7e118deadcd59f5070072c1670b - Update the existing code to walk the DUH document using JSON references rather than hardcoding an (incorrect) path to the actual register definitions.

I tested this on the PIO block by first making some temporary edits to pretend that there are actual bit fields within the PIO control registers and do some sanity testing that the bit twiddling resulted in the correct values.

---

Example header file:

```c
struct metal_pio;

struct metal_pio_vtable {
    void (*v_pio_odata_data_write)(uint32_t * pio_base, uint32_t data);
    uint32_t (*v_pio_odata_data_read)(uint32_t  *pio_base);
    void (*v_pio_oenable_data_write)(uint32_t * pio_base, uint32_t data);
    uint32_t (*v_pio_oenable_data_read)(uint32_t  *pio_base);
    void (*v_pio_idata_data_write)(uint32_t * pio_base, uint32_t data);
    uint32_t (*v_pio_idata_data_read)(uint32_t  *pio_base);
};

struct metal_pio {
    uint32_t *pio_base;
    struct metal_pio_vtable vtable;
};

void metal_pio_odata_data_write(const struct metal_pio *pio, uint32_t data);
uint32_t metal_pio_odata_data_read(const struct metal_pio *pio);
void metal_pio_oenable_data_write(const struct metal_pio *pio, uint32_t data);
uint32_t metal_pio_oenable_data_read(const struct metal_pio *pio);
void metal_pio_idata_data_write(const struct metal_pio *pio, uint32_t data);
uint32_t metal_pio_idata_data_read(const struct metal_pio *pio);
const struct metal_pio *get_metal_pio(uint8_t index);
```

Example C file (some pieces omitted for brevity):

```c
// Private register field access functions

void pio_odata_data_write(uint32_t *pio_base, uint32_t data)
{
    uintptr_t control_base = (uintptr_t)pio_base;
    volatile uint32_t *register_base = (uint32_t *)(control_base + ((PIO_REGISTER_ODATA_DATA / 32) * 4));
    write_field(register_base, (PIO_REGISTER_ODATA_DATA % 32), PIO_REGISTER_ODATA_DATA_WIDTH, data);
}


uint32_t pio_odata_data_read(uint32_t *pio_base)
{
    uintptr_t control_base = (uintptr_t)pio_base;
    volatile uint32_t *register_base = (uint32_t *)(control_base + ((PIO_REGISTER_ODATA_DATA / 32) * 4));
    return read_field(register_base, (PIO_REGISTER_ODATA_DATA % 32), PIO_REGISTER_ODATA_DATA_WIDTH);
}

// Public register field access functions

void metal_pio_odata_data_write(const struct metal_pio *pio, uint32_t data)
{
    if (pio != NULL)
        pio->vtable.v_pio_odata_data_write(pio->pio_base, data);
}


uint32_t metal_pio_odata_data_read(const struct metal_pio *pio)
{
    if (pio != NULL)
        return pio->vtable.v_pio_odata_data_read(pio->pio_base);
    return (uint32_t)-1;
}
```

### DIFF
--- a/scripts/generate_drivers.py
+++ b/scripts/generate_drivers.py
@@ -9,27 +9,10 @@ from dataclasses import dataclass
 from pathlib import Path
 
 import json5
+import jsonref
 
 PlainJSONType = t.Union[dict, list, t.AnyStr, float, bool]
 JSONType = t.Union[PlainJSONType, t.Iterator[PlainJSONType]]
-
-
-def walk(j_obj: JSONType) -> t.Iterator[JSONType]:
-    """
-    Walk a parsed json object, returning inner nodes.
-    This allows the object to be parse in a pipeline like fashion.
-
-    :param j_obj: The object being parsed, or an iterator
-    :return: an iterator of matching objects
-    """
-    if isinstance(j_obj, dict):
-        yield j_obj
-        for v in j_obj.values():
-            yield from walk(v)
-    elif isinstance(j_obj, (list, t.Iterator)):
-        yield j_obj
-        for j in j_obj:
-            yield from walk(j)
 
 
 @dataclass(frozen=True)
@@ -350,9 +333,9 @@ def generate_metal_dev_drv(vendor, device, index, reglist):
 # Support for parsing duh file
 # ###
 
-def walkfile_j5(f_name: str) -> JSONType:
-    "Returns iterator over named json5 file"
-    return walk(json5.load(open(f_name)))
+def load_json5_with_refs(f_name: str) -> JSONType:
+    with open(f_name) as fp:
+        return jsonref.JsonRef.replace_refs(json5.load(fp))
 
 
 ###
@@ -413,21 +396,15 @@ def main():
     m_dir_path = args.metal_dir
     overwrite_existing = args.overwrite_existing
 
-    duh_info = json5.load(open(args.duh_document))
+    duh_info = load_json5_with_refs(args.duh_document)
 
     # ###
     # process pSchema (in duh document) to create symbol table
     # ###
-    p = walkfile_j5(args.duh_document)
-    p = filter(lambda x: 'pSchema' in x, p)
-    p = map(lambda x: x['pSchema'], p)
-    p = filter(lambda x: 'properties' in x, p)
-    p = map(lambda x: x['properties'], p)
-
-    duh_symbol_table = {}
-
-    for prop in p:
-        duh_symbol_table.update(**prop)
+    if 'pSchema' in duh_info['component']:
+        duh_symbol_table = duh_info['component']['pSchema']['properties']
+    else:
+        duh_symbol_table = {}
 
     # ###
     # process register info from duh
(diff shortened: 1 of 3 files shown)


richardxia:
Also, I would advise viewing the diff with whitespace changes hidden, since I had to add an extra level of indentation for a nested `for` loop that iterated over the register fields.

richardxia:
Actually, hold up, I just realized that I basically inlined the field offset definitions directly into the driver, but I should probably be reading them from the `#define`s from the generated header.

However, upon further inspection, I realized that those `#define`s are somewhat wrong, and that's because the `duh-scala` generation is actually a bit off.

richardxia:
OK, new commit https://github.com/sifive/api-generator-sifive/pull/31/commits/2da8a6c630eda8379251f85b8fc40105a77afc73 which implements the read/write helper functions in terms of the `#define`s that come from the generated header files.

scripts/generate_drivers.py
```diff
@@ -40,15 +37,22 @@ class Register:
...
+        name: str,
+        offset: int,
+        width: int,
+        fields: t.List[RegisterField],
+    ) -> "Register":
         if width not in (8, 16, 32, 64):
             raise Exception(f'Invalid register width {width}, for register '
                             f'{name}.\n'
                             f'Width should be not 8, 16, 32, or 64.\n'
                             f'Please fix the register width in DUH document.')
```
sistillson, line 54: Do we want to keep this? It proved to be a problem with header generation

scripts/generate_drivers.py
```diff
@@ -186,10 +197,43 @@ def generate_metal_dev_hdr(vendor, device, index, reglist):
...
+        uint32_t field_offset,
+        uint32_t field_width,
+        uint32_t field_data
+    ) {
+        const uint32_t shifted_field_data = field_data << field_offset;
+        const uint32_t mask = (field_width == 32) ? 0xffffffff : ((1 << field_width) - 1) << field_offset;
+        const uint32_t original_data = *register_base;
+
+        const uint32_t cleared_data = original_data & (~mask);
+        const uint32_t new_data = cleared_data | shifted_field_data;
```
sistillson, line 214: Nice to see the younger generation understanding C bit twiddling.

richardxia: C is still taught in schools!

I will admit, however, that functional programming has poisoned my brain and now I try declaring as many variables as possible as `const` in C.

scripts/generate_drivers.py
richardxia: I was thinking about it while looking over the code, but I decided that I wanted to hold off from doing that for now, since I've already touched a significant number of lines of code in this PR. I think in the common case, that assertion will still hold, and this is a script that's only meant to be run manually anyway, so we don't necessarily need to handle all cases.

richardxia:
I'm circling back on PRs submitted before the holidays. @sistillson, did you have any other comments here, or does this LGTY for merging?

sistillson (approved):
lgtm

PR #34, opened 2020-01-10
Fixed issue with for design with no interrupts

sistillson:
fix generate_header.py to correctly handle a design with no interrupts.


scripts/generate_header.py
```diff
@@ -409,7 +410,10 @@ def main() -> int:
     for index, dev_om in devices_om:
         reglist = find_registers(dev_om)
         intlist = find_interrupts(dev_om, device)
-        base_int = min(i.number for i in intlist)
+        if len(intlist) == 0:
+            base_int = None
+        else:
+            base_int = min(i.number for i in intlist)
```
richardxia: This can be replaced with:

```python
base_int = min(i.number for i in intlint, default=None)
```

The `default` keyword argument allows you to specify a value if the iterable is empty. See https://docs.python.org/3/library/functions.html#min

Could you write a short description of how this was tested? Just want to make sure that this covers everything that needs to be changed if there are no interrupts.

sistillson: fixed, added explicit test to show how I tested it.

scripts/test_object_models/no_interrupts_test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# This is a basic command to test that generate header works with no interrupts defined.
+# Running this file shouldn't raise an exception
+
+../generate_header.py -o no_interrupts.json --vendor sifive -D pio -b no_interrupts -x
```
richardxia: Thanks for adding a test! However, I'd like to request several things:

1. Can you use the long form of the command line flags here? It makes it much easier for someone who is unfamiliar with the short form flags to have the verbose names spelled out. In general, for all scripts, we should invoke commands with the long form of their flags.
2. Can you make this script either independent of the current working directory or print a more helpful message if it is not being run from the expected directory? I've found that it's frustrating to run a script that expects you to be in a specific directory when you invoke it but without it spelling out exactly _where_ it wants you to be, and it's better to either make the script completely agnostic of your working directory or to have it tell you exactly where you should be.
3. Can you make this script exit with a non-zero exit code if it fails? Printing out a pass/fail message is fine, but returning 0 when the tests fail is not.

richardxia (changes requested):

scripts/test_object_models/no_interrupts_test.sh
sistillson: fixed, although there is no easy way to make it directory agnostic.

richardxia: > fixed, although there is no easy way to make it directory agnostic.

Actually, I think there's an easy way to do this, although I think you might need to run on `bash` and not something emulating a Bourne shell:

```bash
dir_name=$(cd "$(dirname "$0")" ; pwd -P)
```

Once you have that, you can `cd` into that directory before invoking everything else.

richardxia (approved):
LGTM!

PR #35, opened 2020-01-13
add vivado resources to bitstream plans

erikdanie:
subscribe to vivadoblackboxresources and add them to the bitstream plan's resource field

### DIFF
--- a/top.wake
+++ b/top.wake
@@ -34,6 +34,7 @@ global def bitstreamTarget board dut buildRoot =
   def resources =
     subscribe vivadoBlackBoxResources
     | mapPartial (_ dut)
+    | map (sortBy (_<*_))
     | flatten
 
   def topModule = dut.getDUTTopModule
--- a/top.wake
+++ b/top.wake
@@ -31,6 +31,10 @@ global def bitstreamTarget board dut buildRoot =
     | map (_ Unit)
     | flatten
     | (dut.getDUTVsrcs ++ _)
+  def resources =
+    subscribe vivadoBlackBoxResources
+    | mapPartial (_ dut)
+    | flatten
 
   def topModule = dut.getDUTTopModule
 
@@ -41,6 +45,7 @@ global def bitstreamTarget board dut buildRoot =
 
   makeBitstreamPlan name vendor board topModule buildRoot.getBuildTreeMCSDir
   | setBitstreamPlanVsrcs vsrcs
+  | editBitstreamPlanResources(resources ++ _)
   | setBitstreamPlanVisibleFiles visibleFiles
   | setBitstreamPlanTclFiles tclFiles
   | makeBitstream


top.wake
```diff
@@ -31,6 +31,10 @@ global def bitstreamTarget board dut buildRoot =
     | map (_ Unit)
     | flatten
     | (dut.getDUTVsrcs ++ _)
+  def resources =
+    subscribe vivadoBlackBoxResources
+    | mapPartial (_ dut)
+    | flatten
```
richardxia, line 38: For reproducibility/determinism purposes, should we sort this? (I guess that might apply to everything else that comes from a `subscribe` as well.)

richardxia (approved):
LGTM!

top.wake
erikdanie: sorry, sort what?

richardxia: Sorting the list of resources. `subscribe` will pull down the list of all `published` things in an arbitrary order. For deterministic builds, we'd probably want this list to always come in the same order.

erikdanie: Ok I had expected to get the subscribed things in random order, in this case I don't believe that they get used in a deterministic way. Even if they did need to be in a certain order, wouldnt that mean I need to know that order when I subscribe? and doesnt that defeat the point of pub/sub?

richardxia: pub/sub is a very generic, low-level language construct, and I think its primary value is in the ability to have many, distributed Wake files publish/"register" with a global topic. Without pub/sub, it would be difficult to implement cases like this, where we may not know all the potential blackbox resources in this file, since others could have been created in independent repos.

I'm not saying that we need to specify things in a "certain" order, but that for reproducibility purposes, we make sure that we always produce the same ordering given the same inputs. For example, if these get serialized into a single string that gets sent to a `Plan`, then the `Plan` will not cache previous runs if the string includes the paths in a different order, even if it should in theory by safe.

erikdanie: Ah ok, that makes perfect sense. So now my question is, how do I sort it? A quick search shows a bunch of things like this:`| sortBy (_<*_)`

richardxia: What is the exact datatype of `resources` here? Are they Paths or Strings, or possibly something else? I don't have all the boolean operators in Wake memorized, but I usually refer to https://sifive.github.io/wake/, which has the types annotated on cursor hover.

erikdanie: ~~Whoops, I'm dumb, I already did this right above the code here, but forgot because I named them slightly different things. Closing this, but will add a sort to that code~~

Nevermind, see below

erikdanie:
Actually I am reopening, as I realize there are 2 different things here, one is adding a file known to git to the vsrcs that get passed to vivado, the other is setting dut comile resources that are used for environment packages

PR #36, opened 2020-01-14
added correct formatting for integer constants larger that 32 bits

sistillson:
If putting in the base address above the 32 bit limit, append ULL to the constant.

Also, added simple test.


sistillson:
As a side note, having tests makes merging super easy if it works.

scripts/generate_header.py
```diff
@@ -155,6 +155,21 @@ class DeviceBase:
...
+    """
+    Takes a positive python integer and return a string that can
+    be used as a constant in C. This means that a L will be prepended
+    if the value is greater than 32 bits.
+
+    :param an_int: an integer value to be written to a c file
+    :return: a string that can be used as a c constant
+    """
+    if an_int > 0xFFFFFFFF:
+        return hex(an_int) + 'ULL'
```
richardxia: Would it be harmful to just always append the `ULL` suffix to the integer literals, even if it's smaller than 32 bits? That may simplify the logic (by obviating most of it).

Also, if you _don't_ unconditionally add the `ULL`, can you at least always add the `U` in the < 2^32 case? It would be really confusing to have it be signed in the case that it is smaller than 2^32, especially since that is a much more likely case.

scripts/generate_header.py
```diff
@@ -155,6 +155,21 @@ class DeviceBase:
 
 # sub templates
 # generate sub parts of template
+
+def integer_from_python_to_c(an_int: int) -> str:
+    """
+    Takes a positive python integer and return a string that can
+    be used as a constant in C. This means that a L will be prepended
+    if the value is greater than 32 bits.
```
richardxia: If you do decide to keep this function, then can you make the following changes to the docstring?

- The word "constant" should be replaced with "literal"
- The word "prepended" should be replaced with "appended"
- "L" should be replaced with "LL", since `long` is a different type than `long long` in C and we should make sure we are precise here.

richardxia (changes requested):
Thanks for adding the test case.

However, can you double check that this actually fixes the error encountered in block-pio-sifive? You should be able to reproduce that case if you modify the generated `pio.scala` to place the device at a very high address.

scripts/generate_header.py
sistillson: fixed. always appends LL, if greater than/equal to zero, appends ULL

Yes, I know zero is zero regardless of sign, but it felt better for some reason to call zero unsigned.

scripts/generate_header.py
sistillson: It is good practice, but doesn't fix the error. I did fix that by changing a u32_t to a uintptr_t in block-pio-sifive/drivers/metal/sifive_pio.c

Sloppiness on my part, assuming that pio was always 32 bit.

scripts/test_object_models/run_tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# Short script to run all the tests
+
+for i in *_test.sh ; do
+    echo ${i}
+    ./$i
+    echo
```
richardxia, line 10: If we're actually going to run multiple tests, we should make sure that this script exits with non-zero status if _any_ of the tests fail.

Also, perhaps we should look at some language-agnostic testing framework. In wit we've also implemented a custom sh-based, testing framework: https://github.com/sifive/wit/tree/0.12.x/t

cc @jackkoenig @mmjconolly

richardxia (approved):

PR #38, opened 2020-01-17
set filename option for rocket-chip generator

albertchen-sifive:
bump rocket-chip to this commit https://github.com/chipsalliance/rocket-chip/pull/2049/commits/43e2c8e776e1bde2e78914a08fe8963ca42ff1f0

set the base filename for rocket-chip outputs so we can avoid `File name too long` errors


erikdanie (approved):
LGTM

richardxia (approved):
LGTM!

PR #41, opened 2020-02-05
Use new Freedom Metal Python-based tools

nategraff-sifive:
Keeping this PR as a draft until the various dependencies of this PR get rolled up into their respective master branches.

In the meantime, this is my take on converting this repository from using the freedom-devicetree-tools based flow to the new python tools. Namely, instead of doing a "fixup", this uses the [devicetree-overlay-generator](https://github.com/sifive/devicetree-overlay-generator) to create a new top-level Devicetree. This overlay generator allows us to dramatically simplify the following tools (especially the linker script generator).

I haven't yet tried to run the new wake scripts (I'm not sure how to drive this repository) but the wake type-checks so it solves my problem at least.

These changes are necessary because we're [replacing the freedom-devicetree-tools implementation of most of the tools the old wake rules used](https://github.com/sifive/freedom-devicetree-tools/pull/204).

### DIFF
--- a/program/freedom-metal-program.wake
+++ b/program/freedom-metal-program.wake
@@ -4,6 +4,18 @@ tuple FreedomMakeAttributes =
   global MABI:   String
   global CMODEL: String
 
+#########################################################
+# runFreedomMakeAttributesGenerator
+#
+# Wraps the esdk-settings-generator and produces both
+# the Makefile fragment for augmenting the build of
+# Freedom Metal, but extracts the ABI, arch, and code
+# model to pass to the compiler.
+#
+# Arguments:
+#  - options: created by makeESDKSettingsGeneratorOptions
+#
+#########################################################
 global def runFreedomMakeAttributesGenerator options =
   def attributes =
     options
@@ -38,7 +50,7 @@ global def makeFreedomBSP coreDTS customOverlay type prefix =
     match customOverlay
       Some overlay =
         # Add coreDTS to the include list in the overlay
-        def coreFilename = extract `.*/(.*)` coreDTS.getPathName | head | getOrElse coreDTS.getPathName
+        def coreFilename = relative "{designDTSName}/.." coreDTS.getPathName
         def list = coreFilename, Nil ++ overlay.getDevicetreeCustomOverlayIncludes
         overlay
         | setDevicetreeCustomOverlayIncludes list
@@ -71,17 +83,15 @@ global def makeFreedomBSP coreDTS customOverlay type prefix =
     makeESDKSettingsGeneratorOptions designDTS (coreDTS, Nil) type "{prefix}.mk"
     | runFreedomMakeAttributesGenerator
 
-  match attributesResult
-    Pass attributes =
-      Pass (
-        FreedomBSP
-        linkerScript
-        header.getFreedomMetalHeaderGeneratorOutputsHeader
-        header.getFreedomMetalHeaderGeneratorOutputsInlineHeader
-        platformHeader
-        attributes
-      )
-    Fail err = Fail err
+  attributesResult |
+  rmap (\attributes
+  FreedomBSP
+  linkerScript
+  header.getFreedomMetalHeaderGeneratorOutputsHeader
+  header.getFreedomMetalHeaderGeneratorOutputsInlineHeader
+  platformHeader
+  attributes
+  )
 
 global def generateMetalFromDts dts machineName customOverlay type outputDir =
   def prefix = "{outputDir}/{machineName}"
@@ -164,9 +174,7 @@ global def freedomMetalDUTProgramCompiler =
       | distinctBy (scmp _.getPathName _.getPathName)
 
     def outputDir = programCompileOptions.getProgramCompileOptionsOutputDir
-    def type = match dut.getDUTFPGABoardOpt
-      None     = "rtl"
-      Some opt = opt
+    def type = dut.getDUTFPGABoardOpt | getOrElse "rtl"
     def installDir = "build/freedom-metal/{dut.getDUTName}"
     def metalInstall =
       generateMetalFromDts dut.getRocketChipDUTDTS dut.getDUTName dut.getDUTCustomOverlay type installDir
--- a/program/freedom-metal-program.wake
+++ b/program/freedom-metal-program.wake
@@ -1,127 +1,79 @@
-# takes a device tree property name and a subnode type and injects a reference
-# to the first occurance of that subnode type to the chosen node
-def addChosenRef name socSubNode inDTSContents =
-  def subnodeOpt =
-    match "({socSubNode}@[a-fA-F0-9]+)".stringToRegExp
-      Pass regex =
-        def nodeRegex = regExpCat (`^\t\t`, `(?:L[a-zA-Z0-9_]+:\s)?`, regex, `\s{$`, Nil)
-        inDTSContents
-        | map nodeRegex.extract
+tuple FreedomMakeAttributes =
+  global File:   Path
+  global MARCH:  String
+  global MABI:   String
+  global CMODEL: String
+
+global def runFreedomMakeAttributesGenerator options =
+  def attributes =
+    options
+    | runESDKSettingsGenerator
+    | getJobOutput
+
+  def rlines = read attributes | rmap `\n`.tokenize
+  match rlines
+    Pass lines =
+      def getAttribute attribute =
+        def regex = regExpCat (`^`, attribute, ` = (.*)$`, Nil)
+        lines
+        | map regex.extract
         | flatten
         | head
-        | omap ("/soc/{_}")
-      Fail _ = None
+      match `RISCV_ABI`.getAttribute `RISCV_ARCH`.getAttribute `RISCV_CMODEL`.getAttribute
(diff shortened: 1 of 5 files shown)


program/freedom-metal-program.wake
```diff
@@ -1,127 +1,79 @@
...
+
+tuple FreedomBSP =
+  global LinkerScript:   Path
+  global Header:         Path
+  global InlineHeader:   Path
+  global PlatformHeader: Path
+  global Attributes:     FreedomMakeAttributes
+
+global def makeFreedomBSP coreDTS prefix =
+  def type = "rtl"
```
nategraff-sifive: The overlay generator (and attributes/esdk settings generator, but it's less critical here) needs a `type` argument from the list of
- rtl
- arty
- vc707
- qemu
- hifive
- spike

This tells the overlay generator which heuristics to follow when picking which memories to use and more. Do we want to expose this as a parameter to `makeFreedomBSP`, and if so, how should that me made available for consumers of `api-generator-sifive`?

nategraff-sifive: Pushed a commit to pass in FPGABoardOpt if it isn't None, otherwise default to "rtl".

nategraff-sifive:
Now that api-languages-sifive and the various new Python tools are happily in their respective masters, I'm ready for review here.

However, **Do Not Merge** until freedom-devicetree-tools is merged to `master` and bumped here, since those changes are still on their own branch.

There are a few main changes to this API that consumers of this API need to know to drive the generation of Freedom Metal programs:

### `DUT.FPGABoardOpts`

The Devicetree overlay generator and makeattributes generator need to know what "type" of target they're trying to generate for. If `FPGABoardOpts` isn't specified, they will expect to find a target for RTL simulation. To generate for Arty or VC707 FPGAs, you'll need to provide a string to `FPGABoardOpts` that has "arty" or "vc707" as a substring.

### `DUT.CustomOverlay`

If a project is using api-generator-sifive to generate Freedom Metal programs for an SoC, odds are the Devicetree overlay generator is going to look at the generated Devicetree for the target and projectile vomit because it doesn't fit any of the heuristics we use. In that case, you're going to want to use the [customOverlay API](https://github.com/sifive/devicetree-overlay-generator/blob/master/customOverlay.wake) to create an explicit description of how you want Freedom Metal to use the generated Devicetree and provide that to the `CustomOverlay` field of the `DUT` tuple.

### `TestProgramPlan`

The above changes replace the use of `Entry`, `Rom`, `Ram`, and `StdoutPath` fields in the `TestProgramPlan` tuple. Those fields are now controlled by the Devicetree Overlay Generator or otherwise supplied to the customOverlay API.

program/freedom-metal-program.wake
```diff
@@ -1,131 +1,91 @@
...
+
+global def makeFreedomBSP coreDTS customOverlay type prefix =
+  # Generate the overlay
+  def designDTSName = "{prefix}-design.dts"
+  def designDTS =
+    match customOverlay
+      Some overlay =
+        # Add coreDTS to the include list in the overlay
+        def coreFilename = extract `.*/(.*)` coreDTS.getPathName | head | getOrElse coreDTS.getPathName
+        def list = coreFilename, Nil ++ overlay.getDevicetreeCustomOverlayIncludes
```
albertchen-sifive: when `coreDTS` and `designDTSName` are in different directories it fails to read the included dts. I think `coreFilename` should be `relative "{designDTSName}/.." coreDTS.getPathName`?

program/freedom-metal-program.wake
```diff
@@ -1,131 +1,91 @@
...
-        def nodeRegex = regExpCat (`^\t\t`, `(?:L[a-zA-Z0-9_]+:\s)?`, regex, `\s{$`, Nil)
-        inDTSContents
-        | map nodeRegex.extract
+tuple FreedomMakeAttributes =
+  global File:   Path
+  global MARCH:  String
+  global MABI:   String
+  global CMODEL: String
+
+global def runFreedomMakeAttributesGenerator options =
```
richardxia, line 19: Can you document the type of the argument to this function? It's not entirely obvious to me from looking at this, and from quickly looking at the code here, I believe that it's a data type that is defined in freedom-e-sdk.

program/freedom-metal-program.wake
```diff
@@ -1,131 +1,91 @@
...
+    Pass attributes =
+      Pass (
+        FreedomBSP
+        linkerScript
+        header.getFreedomMetalHeaderGeneratorOutputsHeader
+        header.getFreedomMetalHeaderGeneratorOutputsInlineHeader
+        platformHeader
+        attributes
+      )
+    Fail err = Fail err
```
richardxia: FYI, if you find yourself doing this pattern of `match` on a `Result` and just passing through the `Fail` case a lot, you can replace it with [`rmap`](https://github.com/sifive/wake/blob/2384a4e456ab7f8104e0030830f7c71f517223de/share/wake/lib/core/result.wake#L40-L42) or [`rmapPass`](https://github.com/sifive/wake/blob/2384a4e456ab7f8104e0030830f7c71f517223de/share/wake/lib/core/result.wake#L44-L48), which are like `map` and `mapFlat` for `Result` types, respectively. In addition to not having to manually write out the failure passthrough, it helps avoid deeply nesting code.

For example, in this case, you could rewrite it as:

```wake
attributesResult
| rmap (\attributes
  FreedomBSP
  linkerScript
  header.getFreedomMetalHeaderGeneratorOutputsHeader
  header.getFreedomMetalHeaderGeneratorOutputsInlineHeader
  platformHeader
  attributes
)
```

program/freedom-metal-program.wake
```diff
@@ -204,15 +164,12 @@ global def freedomMetalDUTProgramCompiler =
       | distinctBy (scmp _.getPathName _.getPathName)
 
     def outputDir = programCompileOptions.getProgramCompileOptionsOutputDir
+    def type = match dut.getDUTFPGABoardOpt
+      None     = "rtl"
+      Some opt = opt
```
richardxia: This can be replaced with the more idiomatic [`getOrElse`](https://github.com/sifive/wake/blob/2384a4e456ab7f8104e0030830f7c71f517223de/share/wake/lib/core/option.wake#L30):

```wake
def type = dut.getDUTFPGABoardOpt | getOrElse "rtl"
```

richardxia (commented):
I think we'll want to make sure it's clear how consumers of the old API can upgrade to the new API. Do `block-pio-sifive` or any of our internal repositories need to be updated to match these changes?

nategraff-sifive:
Bumping this, I've added edits based on review. Does anyone else have feedback or can we get this merged soon?

wit-manifest.json
```diff
@@ -5,7 +5,7 @@
         "source": "git@github.com:sifive/freedom-metal.git"
     },
     {
-        "commit": "89e427e15a403add4eb837f1466f0be9078049f0",
+        "commit": "379db09a9f6cea593357e966788791333ead53c5",
```
nategraff-sifive: Just to be clear, this commit hash points at an unmerged branch. Once approved I can roll up a series of PRs, bump this hash, and then this PR can be merged.

albertchen-sifive (approved):
LGTM

richardxia (approved):

nategraff-sifive:
Ready for merge!

richardxia:
Hey, sorry for the delay; I've been trying to run this commit against the latest master of https://github.com/sifive/block-pio-sifive, just so I can see what the upgrading steps will be like, but I realized that I am not super familiar with what exactly needs to be changed. I'm getting the following error, which I believe implies that I need to set the test memory/boot memory via the new CustomOverlay method, but I need to track down exactly where that is:

```
$ wake runSim pioDUT
'<ivycache>' ivycache
WARNING: Could not parse DTS JSON: _root_.sifive.skeleton.DefaultConfig_sifive.skeleton.WithSimUART_sifive.skeleton.WithTestFinisher_sifive.blocks.pio.WithpioTop.json file not found
python3.7 -B devicetree-overlay-generator/generate_overlay.py -o build/freedom-metal/pioDUT/pioDUT-design.dts -t rtl --rename-include pioDUT.dts build/api-generator-sifive/pioDUT/firrtl/pioDUT.dts
devicetree-overlay-generator/generate_overlay.py: Unable to determine test bench reset vector

===== TEST RESULTS =====
- test: demo
  msg: FAILED: freedom-metal-dut-program-compiler: failed to compile binary

Fail "Some tests failed"
```

Note that I think this has nothing to do with your PR, and more to do with how I need to understand how https://github.com/sifive/soc-testsocket-sifive interacts with api-generator-sifive and block-pio-sifive.

nategraff-sifive:
soc-testsocket-sifive uses a RocketChipDUTPlan. Adding support for these tools just means adding a few fields to the `RocketChipDUTPlan` tuple and then mapping those through when it gets converted to a `DUT`. Pushed the necessary changes here, and [soc-testsocket-sifive/build.wake](https://github.com/sifive/soc-testsocket-sifive/blob/master/build.wake) can be modified in the same way that @albertchen-sifive has already shown works internally.

albertchen-sifive (approved):
tested that new changes work with block-pio-sifive locally

richardxia (approved):
LGTM!

nategraff-sifive:
Ready for someone to hit the merge button (I don't have privilege on this repository).

PR #42, opened 2020-02-11
add runVerilator that runs Verilator without Make

albertchen-sifive:
some updates for upcoming build json PR:
- add a `Visible` field to `VCSCompilePlan` for adding extra visible files
- factor Verilator compile job from `doVerilatorCompile` into a new global function `runVerilator`

### DIFF
--- a/sim/verilator.wake
+++ b/sim/verilator.wake
@@ -8,6 +8,7 @@ tuple VerilatorCompilePlan =
   global Main:          Path
   global Prefix:        String
   global TopModule:     String
+  global Visible:       List Path
   global Resources:     List String
   global OutputDir:     String
 
@@ -24,6 +25,7 @@ global def makeVerilatorCompilePlan sourceFiles main topModule outputDir =
   main           # Main
   "V{topModule}" # Prefix
   topModule      # TopModule
+  Nil            # Visible
   defaultVerilatorResources # Resources
   outputDir      # OutputDir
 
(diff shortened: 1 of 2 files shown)


sim/verilator.wake
```diff
@@ -64,51 +66,75 @@ def compileCmdlineArgs options =
...
 
 def compileInputs options =
-  def outputDir    = options.getVerilatorCompilePlanOutputDir
-  def includeDirs  = options.getVerilatorCompilePlanIncludeDirs
-  def srcFiles     = options.getVerilatorCompilePlanSourceFiles
-
+  def outputDir = options.getVerilatorCompilePlanOutputDir
+  def includeDirs = options.getVerilatorCompilePlanIncludeDirs
+  def srcFiles = options.getVerilatorCompilePlanSourceFiles
+  def visible = options.getVerilatorCompilePlanVisible
```
richardxia, line 72: I figure I might as well ask this code style question here: What are your thoughts on vertically aligning assignment statements? I see here that you have switched away from vertically aligning the `=` signs and instead to just having a single space on both sides of the `=` sign. My personal preference is to not vertically align assignment statements, since I feel that too often you will add a variable with a longer name and either have to touch all the other lines or not have everything line up anymore. I'm fine with either approach, but I'd rather that we all have a consistent style instead of changing back and forth.

sim/verilator.wake
```diff
@@ -64,51 +66,75 @@ def compileCmdlineArgs options =
...
   def includeSources = map (sources _ `.*`) includeDirs | flatten
 
-  source verilatorWrapperScript, mkdir outputDir, includeSources ++ srcFiles# ++ includeDirs
+  source verilatorWrapperScript, mkdir outputDir, includeSources ++ srcFiles ++ visible
 
-global def doVerilatorCompile options =
+def filterVerilatorOutputs all =
+  filter (\f !matches `.*\.(d|dat|tree|dot|a)` f) all
+
+global def runVerilator options =
```
richardxia, line 81: Can you add (at least a short, one-sentence) comment describing this function and `doVerilatorCompile`, and specifically describing what is different about them? Within the context of this PR, I understand, but for someone just digging through the codebase, it's less clear what the difference is between the two.

richardxia (approved):
LGTM!

sim/verilator.wake
albertchen-sifive: yeah I used to like aligning things but now I feel like it's more trouble than it's worth so I'm switching things back.

PR #43, opened 2020-02-16
Correct some typos in no_interrupts.json

felixonmars:


richardxia (approved):
Thanks for finding these! Although since this file is automatically generated, we should fix them in the source repository in rocket-chip as well: https://github.com/chipsalliance/rocket-chip/blob/9ad93324fa97be9769c6c4cee0e383e0ecf84be0/src/main/scala/devices/debug/Debug.scala#L1254

I'll accept this patch for now and submit a patch against rocket-chip to correct the typos.

PR #44, opened 2020-02-25
vcu118 support checkin

erikdanie:
Deal with either one or multiple Mcs files. I'm not sure how to deal with the getOrElse since I don't have the name anymore, so I just removed it. If anyone has better ideas I am open to that too

### DIFF
--- a/fpga/mcs.wake
+++ b/fpga/mcs.wake
@@ -123,13 +123,20 @@ global def makeMCS plan =
     | setVivadoPlanVisibleFiles (outputDir.mkdir, bitstream, programImgList ++ tclSupportFiles)
     | runVivado
 
-  vivadoMCS
-  | getVivadoOutputAllOutputs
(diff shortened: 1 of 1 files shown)


fpga/mcs.wake
```diff
@@ -118,20 +118,18 @@ global def makeMCS plan =
...
   vivadoMCS
   | getVivadoOutputAllOutputs
-  | head
-  | getOrElse "MCS failed: {outputFile}".makeError.makeBadPath
   | MCSOutputs plan
 
 tuple MCSOutputs =
   Plan_: MCSPlan
-  MCS_: Path
+  MCS_: List Path
```
albertchen-sifive: can we add another field instead of changing the type of an existing field to maintain backwards compatibility? like a new `AllMCS` field that is a list of the `*.mcs` files, and the existing `MCS` field still points to whatever file it was pointing at before

PR #47, opened 2020-03-10
generate_drivers.py: Handle JSON References to local files.

### LINKED ISSUE #45: generate_drivers.py not handling references to local files
If the input DUH document uses a [JSON Reference](https://tools.ietf.org/id/draft-pbryan-zyp-json-ref-03.html) to a local file using a relative file path (e.g. `'./foo.json5'`), then the `generate_drivers.py` script will error out with:

```pytb
jsonref.JsonRefError: ValueError: unknown url type: './foo.json5'
```

It looks like the JSON References library we are using does support customizing the [loader method](https://github.com/gazpachoking/jsonref/blob/066132e527f8115f75bcadfd0eca12f8973a6309/jsonref.py#L80-L81), which gives us an extra hook for customizing how files are loaded, which should allow us use a loader that loads from disk rather than HTTP when the JSON Reference is to a local file.

richardxia:
Fixes #45.

Adds a custom loader that checks a URI for whether it is a web URL or just a local file path and uses the appropriate method for loading the JSON file. The jsonref library will otherwise always attempt to fetch from the internet.

Tested this locally on a DUH document that has a JSON reference to a different file using a relative path.


albertchen-sifive (approved):

drom (commented):
LGTM

PR #49, opened 2020-03-19
Fix DTSJSON Path when base file names are overridden.

richardxia:
As @davidmlw mentioned in https://github.com/sifive/api-generator-sifive/commit/e238d26e0dcc9f8f36edb9dc4fb2ae18cdd3667f#commitcomment-37789577, the DTS JSON path is incorrect if the `RocketChipGeneratorOptions` `BaseFileName` option is set, since the DTS JSON file path has a hardcoded `_root_` that is only used when no option is set.

This fixes the issue in api-generator-sifive.git by deriving the base file name from the BaseFileName. I wrote the logic so that instead of trying a default value when that `BaseFileName` is not set, it just returns a `BadPath`. I think this is somewhat safe to do because in this file, we always set `BaseFileName`, so even though the `BaseFileName` is an option, we can be more confident that it is always be set here.

The ideal solution would be to actually provide the `dtsJSON` path from rocket-chip.git's Wake file, since that's where it has the logic for defaulting to `_root_` in the first place. However, I opted to not do that since that would require making an update in rocket-chip and updating api-generator-sifive to point to it, which does not seem to just work out of the box, because we have no continuous integration set up on this repository.

If I have time, I'll add the `dtsJSON` option to rocket-chip as well so that we can later opportunistically remove the code path I have added here.

Tested locally on a configuration that was previously failing to find the DTS JSON and confirmed that it is in fact being detected now.

### DIFF
--- a/dut/rocket-chip-dut.wake
+++ b/dut/rocket-chip-dut.wake
@@ -56,6 +56,10 @@ global def getRocketChipDUTDTS         = getDUTDTS
 
 def getTestharnessTopModule testharness = (tokenize `\.` testharness).reverse.head
 
+def getWhenFailFn fn = match _
+  Pass a = a
+  Fail f = fn f
+
 global def rocketChipDUTMaker dutPlan userArgs =
   def rocketOutputs = doRocketGenerate dutPlan userArgs
   def firrtlOutputs = doFirrtlCompile dutPlan userArgs rocketOutputs
@@ -72,20 +76,31 @@ global def rocketChipDUTMaker dutPlan userArgs =
   def allOutputs = rocketOutputs.getRocketChipGeneratorOutputsAllOutputs
   def topModuleOpt = dutPlan.getRocketChipDUTPlanTestharness.getTestharnessTopModule
   def dts = rocketOutputs.getRocketChipGeneratorOutputsDTS
+
+  # TODO: This should be something that is captured by
+  # RocketChipGeneratorOutputs in rocket-chip.git, since only the Wake rules in
+  # that repo can know what the final output filenames look like.
+  # This can be replaced once we actually add that in rocket-chip.git.
   def dtsJSON =
-    def fileName =
+    def fileNameOpt =
       rocketOutputs
       | getRocketChipGeneratorOutputsInputOptions
-      | getRocketChipGeneratorOptionsConfigNames
-      | catWith "_"
-      | ("_root_.{_}.json")
-    def fileRegex = regExpCat (`([^/]*/)*`, fileName.quote, Nil)
+      | getRocketChipGeneratorOptionsBaseFileName
+      | omap ("{_}.json")
 
-    rocketOutputs
-    | getRocketChipGeneratorOutputsAllOutputs
-    | find (matches fileRegex _.getPathName)
-    | omap getPairFirst
-    | getOrElse "{fileName} file not found".makeError.makeBadPath
+    def getDTSJSONForFileName fileName =
+      def fileRegex = regExpCat (`([^/]*/)*`, fileName.quote, Nil)
+
+      rocketOutputs
+      | getRocketChipGeneratorOutputsAllOutputs
+      | find (matches fileRegex _.getPathName)
+      | omap getPairFirst
+      | getOrFail "{fileName} file not found".makeError
+
+    fileNameOpt
+    | getOrFail "dtsJSON not available if outputBaseName not explicitly passed to rocket-chip Generator.scala.".makeError
+    | rmapPass getDTSJSONForFileName
+    | getWhenFailFn (_.makeBadPath)
 
   match objectModelResult topModuleOpt
     (Pass objectModel) (Some topModule) =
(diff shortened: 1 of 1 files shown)


dut/rocket-chip-dut.wake
```diff
@@ -72,20 +76,31 @@ global def rocketChipDUTMaker dutPlan userArgs =
...
+      def fileRegex = regExpCat (`([^/]*/)*`, fileName.quote, Nil)
+
+      rocketOutputs
+      | getRocketChipGeneratorOutputsAllOutputs
+      | find (matches fileRegex _.getPathName)
+      | omap getPairFirst
+      | getOrFail "{fileName} file not found".makeError
+
+    fileNameOpt
+    | getOrFail "dtsJSON not available if outputBaseName not explicitly passed to rocket-chip Generator.scala.".makeError
```
albertchen-sifive: if its not specified then wouldn't the filename be `_root_.config1_config2.json` or whatever? or is this intentional so that we don't rely on the this?

dut/rocket-chip-dut.wake
```diff
@@ -72,20 +76,31 @@ global def rocketChipDUTMaker dutPlan userArgs =
...
+      | omap ("{_}.json")
+
+    def getDTSJSONForFileName fileName =
+      def fileRegex = regExpCat (`([^/]*/)*`, fileName.quote, Nil)
+
+      rocketOutputs
+      | getRocketChipGeneratorOutputsAllOutputs
+      | find (matches fileRegex _.getPathName)
+      | omap getPairFirst
+      | getOrFail "{fileName} file not found".makeError
```
albertchen-sifive: I think this is a better pattern. this way you propagate the original error message if the job failed. and also prevent matching on the path name `BAD_PATH`
```
outputs
| findFailFn getPathResult
| rmapPass (
  _
  | find (matches fileRegex _.getPathName)
  | omap getPairFirst
  | getOrFail "{fileName} file not found".makeError
)

**Review by albertchen-sifive — APPROVED** [review]

File: dut/rocket-chip-dut.wake (hunk 1 above)
Comment (albertchen-sifive) [reply dut/rocket-chip-dut.wake]: nevermind, just read the PR comment

Comment (richardxia) [reply dut/rocket-chip-dut.wake]: Yeah, I would have wanted to make the changes just in rocket-chip, and I'm sort of seeing if I can sneak out that PR now, but there are a lot of things that depend on non-backwards incompatible changes to both api-generator-sifive.git and rocket-chip.git, so I thought that this would be good enough for now.

File: dut/rocket-chip-dut.wake (hunk 2 above)
Comment (richardxia) [reply dut/rocket-chip-dut.wake]: Let me double check that I understand the difference between what I had and what you wrote. In your approach, you first filter the job for all and any failed paths prior to trying to extract the desired path out of the outputs list, which means that we can error out earlier if the whole thing failed. In my approach, the `find` expression would merely fail to match anything because the BadPaths would not have real PathNames, and it would fall through to the "file not found" error. Is this correct?

cc @terpstra as an example of where this pattern can lead to intuitive and awkward results.

I'll go with your approach, @albertchen-sifive.

Comment (albertchen-sifive) [reply dut/rocket-chip-dut.wake]: yes that's correct

Comment (richardxia) [reply dut/rocket-chip-dut.wake]: Might have spent a bit too long figuring out how to write this, but I think I have this factored in a way that I like and that does not rely on nested pipelines: https://github.com/sifive/api-generator-sifive/pull/49/commits/331bbb2a3052214d316632f819c203a60e09b031

File: dut/rocket-chip-dut.wake [hunk 3]
```diff
@@ -72,20 +72,42 @@ global def rocketChipDUTMaker dutPlan userArgs =
...
+      def fileRegex = regExpCat (`([^/]*/)*`, fileName.quote, Nil)
+      outputs
+      | find (matches fileRegex _.getPathName)
+      | omap getPairFirst
+      | getOrElse "{fileName} file not found".makeError.makeBadPath
+
+    match (Pair (getFileName rocketOutputs) (getOrFailOutputs rocketOutputs))
+      Pair (Fail err) _ = err.makeBadPath
+      Pair _ (Fail err) = err.makeBadPath
+      Pair (Pass fileName) (Pass outputs) = findDTSJSONPath fileName outputs
```
albertchen-sifive: you can match on two separate things without putting them in a `Pair`
```
    match (getFileName rocketOutputs) (getOrFailOutputs rocketOutputs)
      (Fail err) _ = err.makeBadPath
      _ (Fail err) = err.makeBadPath
      (Pass fileName) (Pass outputs) = findDTSJSONPath fileName outputs
```

albertchen-sifive (approved):

dut/rocket-chip-dut.wake
richardxia: Weird, I had tried that earlier but it wasn't working for me. Maybe it was from an earlier version of my match statement when I hadn't changed the arguments to function calls on `rocketOutputs`, and it was getting confused? I think before I had something like:

```wake
    match fileNameOpt maybeRocketOutputs
      (Fail err) _ = err.makeBadPath
      _ (Fail err) = err.makeBadPath
      (Pass fileName) (Pass outputs) = findDTSJSONPath fileName outputs
```

In that case, I'm guessing that I would have to parenthesize each item so that it doesn't think that I'm attempting to call `fileNameOpt` on `maybeRocketOutputs`?

Anyway, thanks, I'll update this.

richardxia:
Thanks for the review and Wake style tips! Going to merge this in now.

PR #51, opened 2020-03-20
Added the Verdi includes for compilation

RajeshVaradharajan:


richardxia (approved):
Thanks! I double checked that this works for me.

PR #52, opened 2020-03-20
Install firrtl-generated DTS in BSP directory

nategraff-sifive:


nategraff-sifive:
Requesting review from @richardxia @albertchen-sifive since I can't add them in the sidebar

albertchen-sifive (approved):

nategraff-sifive:
I don't have write privileges, merge at will :)

richardxia (approved):
Thanks!

program/freedom-metal-program.wake
```diff
@@ -43,44 +43,46 @@ tuple FreedomBSP =
   global PlatformHeader: Path
   global Attributes:     FreedomMakeAttributes
 
-global def makeFreedomBSP coreDTS customOverlay type prefix =
+global def makeFreedomBSP coreDTS customOverlay type prefix machineName =
```
richardxia, line 46: Wait, this makes a backwards-incompatible change to this API. Since it's `global`, let me quickly check if we have any other codebases that reference this.

richardxia: OK, this seems to not be used anywhere outside of this file. Is this something we should consider making non-global? If not, then I think we'll want to more gracefully handle forwards-compatibility by wrapping the arguments into `tuples`, for which we can define accessor methods for new fields in an incremental fashion.

PR #53, opened 2020-03-24
Use installAs, not installIn for DTS path.

richardxia:
Refs https://github.com/sifive/api-generator-sifive/pull/52

/cc @nategraff-sifive 

It turns out that the definition of [`installIn`](https://github.com/sifive/wake/blob/e9bd5c3c0131d833c9ba2aeb20344a3808d0be7a/share/wake/lib/system/cp.wake#L32-L33) will essentially concatenate the target directory with the target file's full relative path, meaning that `installIn "foo" (source "bar/baz.json")` will copy the file to `foo/bar/baz.json`, _not_ to `foo/baz.json`.

What we actually want here is to compute the full desired path and use `installAs` to place the file directly there.

I also fixed a related issue with the pure-wake CustomOverlay generator getting the full relative path and inserting that into the `/include/` statement by also getting the base name of the file. I think this is the right place to put this logic, since the "knowledge of the relative paths of the two DTS files" is within this `makeFreedomBSP` function, but please let me know if you think it belongs elsewhere.

I tested this locally by fixing the specific issue I was encountering with the DTS file being copied to the wrong place as well as the custom overlay DTS having a long include path. I also tested this locally in one of our larger, integration repositories internally to make sure that this doesn't cause a regression in the expected DTS behavior.


albertchen-sifive (approved):

richardxia:
I know that @nategraff-sifive can't actually approve/reject these changes, since he doesn't have write access to this repo, but I just want to wait for an unstructured yes/no here before merging.

nategraff-sifive:
Approved :)

richardxia:
Thanks!

PR #54, opened 2020-03-24
Remove unused sim option tuples.

richardxia:
I keep on getting confused by these, and I just realized that they don't seem to be used anywhere, since their constructor is never invoked, ever.

Am I missing something here, or do these actually serve a purpose? My Wake files seem to compile even when these are not there.


albertchen-sifive (approved):

richardxia:
Thanks for the quick review!

PR #56, opened 2020-03-25
Bumped the verilator version from 4.008 to 4.028

RajeshVaradharajan:


sim/verilator.wake
```diff
@@ -13,7 +13,7 @@ tuple VerilatorCompilePlan =
   global Resources:     List String
   global OutputDir:     String
 
-global def defaultVerilatorResources = "verilator/4.008", Nil
+global def defaultVerilatorResources = "verilator/4.024", Nil
```
richardxia:
```suggestion
global def defaultVerilatorResources = "verilator/4.028", Nil
```

Actually, I'm going to change this to 4.028, since that's what we're using in a few places internally, and it'd be nice to make them match.

richardxia:
Thanks for fixing this!

PR #57, opened 2020-03-26
named-args.wake: Privatize named arg functions

richardxia:
Unless if there's a strong reason to leave these as global/public, I've gone ahead and made most of these functions private. I was having a bit of difficulty understanding the semantics of the `dir` argument to `namedArgsToCmdline`, and it was a little bit difficult for me to understand, since I had to chase the call stack following the `dir` argument all the way to `getNamedArgValue`, and then "pop back out" the stack to remember what I was doing.

I added documentation, but rather than documenting all of these functions that take in the `dir` argument, I thought it might be a nicer API to only force callers of `namedArgsToCmdline` to have to understand what `dir` means rather than documenting every single one of these functions.

I did a quick check across several of our internal repositories and I do not see most of these functions being called directly, so it should be safe to make most of them private.

---

On a side note, it feels that the abstraction that we were going for here is a bit off center. I know it'll require a lot of work to migrate, especially given that repositories like https://github.com/sifive/block-pio-sifive contribute their own `NamedArg` instances, but I was wondering if you think it would be better to eventually replace all this with a simpler abstraction that does not attempt to provide an abstraction for choosing between `--` and `+` as the prefix, `=` as an operator, or requiring that any non-string arguments must be paired with a "name". For the latter, I think we've run into trouble when we wanted to pass in a `Path` without having an accompanying name. In addition, we've also run into trouble when we want to mix and match `--` and `+` prefixes, since _both_ are valid to pass into Verilog simulation tools.

### DIFF
--- a/utils/named-arg.wake
+++ b/utils/named-arg.wake
@@ -5,35 +5,35 @@ global data NamedArg =
   NamedArgString  String String
   NamedArgPath    String Path
 
-global def getNamedArgName = match _
+def getNamedArgName = match _
   NamedArg        n   = n
   NamedArgInteger n _ = n
   NamedArgDouble  n _ = n
   NamedArgString  n _ = n
   NamedArgPath    n _ = n
 
-global def getNamedArgValue dir = match _
+def getNamedArgValue dir = match _
   NamedArg        _   = None
   NamedArgInteger _ i = Some (str i)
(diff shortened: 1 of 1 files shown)


richardxia:
Stale.

PR #60, opened 2020-04-07
Stop joining register and field names.

richardxia:
When generating the C header files used in software compilation, we had a piece
of logic that would attempt to eliminate duplicate words that appeared at the
end of a register name and the beginning of a field name. For example, given a
register named `FOO_READ` and a register field named `READ_STROBE`, we would
produce a C macro that would be named `FOO_READ_STROBE` rather than
`FOO_READ_READ_STROBE`.

Although the coalesced version may read a little bit nicer in English, it adds
complexity in how the C macro names are generated, which is an important API
between the drivers and the header file, so it's probably better to not handle
it specially and just concatenate the register name with the field name.

This PR removes the extra bit of logic so that the C macros have simpler, though possibly redundant, names.


mwachs5:
The original intention seems weird (Unless you had a register like "STATE" with only field called "STATE" ...?) So I support this change.

mwachs5 (approved):

erikdanie (approved):

PR #61, opened 2020-04-09
Fix syntax for `target`

richardxia:
It looks like https://github.com/sifive/api-generator-sifive/commit/ef7e4df853dfa9fc47e2a0cce63411ffbebb4061 introduced some invalid syntax for a `target`. The `target` keyword _replaces_ the `def` keyword, so it has the function of binding a value to a name rather than modifying a definition.

This PR changes it to be what I think you wanted it to be. I'm going to merge this without review since I have confirmed that this compiles for me and because this is blocking a different use case where we can't bump to a commit where I had fixed a bug (https://github.com/sifive/api-generator-sifive/commit/1538a86cf354c0979a6a70ebd7dfbf90eb802d67).

Adding @terpstra as a reviewer, since I, too, found the behavior of `target`/`memoize` a bit confusing at first.


terpstra:
I am confused that 'target def' worked. AFAIK, that has never been legal.

richardxia:
`target def` did not work, which is why I made this change.

PR #62, opened 2020-04-27
Handle address blocks in software headers and drivers

richardxia:
One component of https://github.com/sifive/duh-scala/issues/79.

This updates both the `generate_headers.py` and `generate_drivers.py` scripts for the IP onboarding workflow to handle address blocks. My goals were to 1) seamlessly support the notion of the address blocks by including the address block name in the generated C macros and 2) preserve backwards compatibility for existing IP blocks.

What was a little bit challenging is that although `generate_headers.py` derives its information from the Object Model, `generate_drivers.py` derives its information directly from DUH. Although I could automatically detect the address block scenario in the former by checking whether the Object Model actually contains any references to address blocks, the latter was more challenging, since DUH has always had mandatory address blocks for describing registers.

For the best backwards and forwards compatibility, I went with the following rules:

- `generate_headers.py` - Always emit the old-style macros that didn't have the address blocks, but in addition emit the new-style macros if the address block property is present in the Object Model.
- `generate_drivers.py` -  If there are multiple address blocks, always emit the driver with the addresss block name in the macro; otherwise emit the old-style macro. This is overridable with the `--always-include-address-block-in-macros` command-line argument.

The other change that I made was to update `generate_headers.py` to also emit the relative base address of each address block as a set of standalone macros, in case if someone wants to handwrite a driver that references the address block base address. This is probably useful in the case where the address block doesn't actually contain registers, such as when it acts as a memory-like region rather than memory-mapped control registers.

---

I tested this by hacking up the block-pio-sifive flow to add in a second address block, as well as faking out the Object Model with a hand-written Object Model in which I added the OMAddressBlock and other information.

I attempted to test this both with and without my hacked up changes, to make sure that I did indeed preserve backwards compatibility, but I think it'll be hard to know if this works for real until duh-scala is updated to emit the OMAddressBlock information.

### DIFF
--- a/scripts/generate_drivers.py
+++ b/scripts/generate_drivers.py
@@ -17,6 +17,21 @@ JSONType = t.Union[PlainJSONType, t.Iterator[PlainJSONType]]
 
 
 @dataclass(frozen=True)
+class AddressBlock:
+    """
+    A DUH address block.
+
+    Note that for legacy reasons, this does not include the registers in the
+    address block. The previous code in this script primarily operated on
+    registers, and introducing a level of hierarchy for the address blocks
+    would require essentially a rewrite of this entire script.
+    """
+    name: str
+    baseAddress: int
+    range: int
+    width: int
+
+@dataclass(frozen=True)
 class RegisterField:
     """
     Description of a bit field within a register.
@@ -39,6 +54,7 @@ class Register:
     offset: int  # in bytes
     width: int  # in bits
     fields: t.List[RegisterField]
+    address_block: AddressBlock
 
     @classmethod
     def make_register(
@@ -47,13 +63,14 @@ class Register:
         offset: int,
         width: int,
         fields: t.List[RegisterField],
+        address_block: AddressBlock,
     ) -> "Register":
         if width not in (8, 16, 32, 64):
             raise Exception(f'Invalid register width {width}, for register '
                             f'{name}.\n'
                             f'Width should be not 8, 16, 32, or 64.\n'
                             f'Please fix the register width in DUH document.')
-        return cls(name, offset, width, fields)
+        return cls(name, offset, width, fields, address_block)
 
 
 ###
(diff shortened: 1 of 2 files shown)


mwachs5:
On this branch of block-pio-sifive (https://github.com/sifive/block-pio-sifive/tree/duh-79-tests) where I have tweaked the PIO block to have a more interesting memory map, I am getting this error:

```
python3.7 -B api-generator-sifive/scripts/generate_header.py --object-model build/api-generator-sifive/pioDUT/programs/omfile.json --vendor sifive --device pio --bsp-dir build/api-generator-sifive/pioDUT/programs/demo/pio --overwrite-existing
Traceback (most recent call last):
  File "api-generator-sifive/scripts/generate_header.py", line 511, in <module>
    sys.exit(main())
  File "api-generator-sifive/scripts/generate_header.py", line 476, in main
    fields = find_register_fields(dev_om)
  File "api-generator-sifive/scripts/generate_header.py", line 368, in find_register_fields
    for aReg in mr['registerMap']['registerFields']:
KeyError: 'registerMap'
```

Note that this version of the PIO does not always have a registerMap for every memory region. were you expecing this change to handle the case when there was no register map for a memory region?

This is the OM snippet for the region with no registers:

```
, {
      "name" : "pio@2000000000000",
      "description" : "JustMemoryNoRegisters",
      "addressSets" : [ {
        "base" : 844424930131968,
        "mask" : 4095,
        "_types" : [ "OMAddressSet", "OMCompoundType" ]
      } ],
      "permissions" : {
        "readable" : true,
        "writeable" : true,
        "executable" : false,
        "cacheable" : false,
        "atomics" : true,
        "_types" : [ "OMPermissions", "OMCompoundType" ]
      },
      "addressBlocks" : [ ],
      "_types" : [ "OMMemoryRegion", "OMCompoundType" ]
    }, 
```

I can attach the full OM if you'd like

mwachs5:
Hm, I am realizing that none of the addressBlocks are populated in the object model, which is a different issue, so you may want to hold off for now.

richardxia:
> Hm, I am realizing that none of the addressBlocks are populated in the object model, which is a different issue, so you may want to hold off for now.

Understood, although you are correct that the current script does not work for any block that doesn't have any registers within a memory region (e.g. SiFive's SPI Flash peripheral). I had noticed this issue in the past, but haven't prioritized fixing it yet, but it is something that ought to be fixed eventually.

mwachs5:
ok, so I fixed my branch to emit the OM properly and then manually deleted the "no registers" region from the OM and the script ran successfully. The output looks OK.

richardxia:
> ok, so I fixed my branch to emit the OM properly and then manually deleted the "no registers" region from the OM and the script ran successfully. The output looks OK.

OK, that's good to hear, but do you need me to properly handle the case where there are no registers in a region?

PR #64, opened 2020-05-13
Bump scribble-testsocket-sifive to pick up ErrorDevice fix.

richardxia:
Bumps to version of scribble-testsocket-sifive pointed to by https://github.com/sifive/scribble-testsocket-sifive/pull/4.

Note that I am planning to squash-merge that other PR and to tag it as v0.1.4, so the commit hash will change here.

### DIFF
--- a/wit-manifest.json
+++ b/wit-manifest.json
@@ -20,7 +20,8 @@
         "source": "git@github.com:sifive/api-languages-sifive.git"
     },
     {
-        "commit": "c675ef819c022a4871bc7a4db8e40018dd188ea2",
+        "//": "This commit is v0.1.4",
+        "commit": "081e45cb46f65ad6a0e251f357e995d2a795b4c7",
         "name": "scribble-testsocket-sifive",
         "source": "git@github.com:sifive/scribble-testsocket-sifive.git"
     },


albertchen-sifive (approved):

wit-manifest.json
```diff
@@ -20,7 +20,8 @@
         "source": "git@github.com:sifive/api-languages-sifive.git"
     },
     {
-        "commit": "c675ef819c022a4871bc7a4db8e40018dd188ea2",
+        "//": "This commit is v0.1.4",
+        "commit": "081e45cb46f65ad6a0e251f357e995d2a795b4c7",
```
richardxia, line 24: I just updated this to point at the [tagged commit for scribble-testsocket-sifive v0.1.4](https://github.com/sifive/scribble-testsocket-sifive/tree/v0.1.4).

Any thoughts on whether it's useful to include the comment string I just added above? I know it'd probably be better for Wit to just natively support tags/branches, but in the meantime, this at least makes it easier to see as a human what version the commit corresponds to. The downside is that it's easy for that comment to get out of date if we're not disciplined about updating it.

richardxia:
Since we already merged in https://github.com/sifive/soc-testsocket-sifive/pull/20 pointing to this commit., let me just merge this in with a merge commit as to preserve the original commit.

PR #70, opened 2020-06-24
api for synth only fpga ci target

erikdanie:
pretty much copies of the main fpga paths that call different scripts. I thought about having these follow the same path but it seemed more difficult than it was worth, especially since these are mostly just used as a quick CI check to make sure there aren't any glaring issues

### DIFF
--- a/fpga/mcs.wake
+++ b/fpga/mcs.wake
@@ -19,6 +19,61 @@ def vc707migHook =
   def addResources = "vivado/vc7071gmig/1.0"
   makeBlackBoxHook name addResources
 
+global target fpgaSynthOnly plan =
+  def name = plan.getBitstreamPlanName
+  def resources = plan.getBitstreamPlanResources
+  def outputDir = plan.getBitstreamPlanOutputDir
+  def designTopModule = plan.getBitstreamPlanTopModule
+  def vendor = plan.getBitstreamPlanVendor
+  def board = plan.getBitstreamPlanBoard
+  def visibleFiles = plan.getBitstreamPlanVisibleFiles
+
+  def vivadoSynth =
+    def sourceTcl = source "fpga-shells/{vendor}/common/tcl/synth-only.tcl"
+    def tclFiles = plan.getBitstreamPlanTclFiles
+    def vsrcs = plan.getBitstreamPlanVsrcs
+    def runDir = outputDir
+
+    def ffile =
+      map (relative runDir _.getPathName) vsrcs
+      | catWith "\n"
+      | write "{runDir}/verilog.F"
+
+    def tclArgs =
+      def tclsString =
+        tclFiles
+        | map (relative runDir _.getPathName)
+        | ("{catWith "  " _}")
+      "-top-module",     designTopModule,
+      "-board",          board,
+      "-F",              (relative runDir ffile.getPathName),
+      "-ip-vivado-tcls", tclsString,
+      "-env-var-srcs",   "WAKE_IP_RESOURCE_FILES",
+      Nil
+
+    def fnOutputs _ = files outputDir `.*\.bit`
+
+    def tclSupportFiles =
+      (sources "fpga-shells/{vendor}/{board}/tcl" `.*\.tcl`)
+      ++ (sources "fpga-shells/{vendor}/common/tcl/" `.*\.tcl`)
+      ++ (sources "fpga-shells/{vendor}/{board}/constraints/" `.*\.xdc`)
+
+    def allVisibleFiles =
+      (outputDir.mkdir, ffile, Nil)
+      ++ visibleFiles
+      ++ tclSupportFiles
+      ++ vsrcs
+      ++ tclFiles
+
+    makeVivadoPlan sourceTcl outputDir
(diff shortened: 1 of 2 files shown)


fpga/mcs.wake
```diff
@@ -19,6 +19,61 @@ def vc707migHook =
   def addResources = "vivado/vc7071gmig/1.0"
   makeBlackBoxHook name addResources
 
+global target fpgaSynthOnly plan =
```
richardxia: Can we make this function name a "verb" so that it's clearer that it "does" something? How about "synthesizeFPGA", or "runFPGASynthesis"?

Also, could you document the type signature of this target? Especially the return type, since it's not obvious from reading the definition.

fpga/mcs.wake
```diff
@@ -19,6 +19,61 @@ def vc707migHook =
...
+  def name = plan.getBitstreamPlanName
+  def resources = plan.getBitstreamPlanResources
+  def outputDir = plan.getBitstreamPlanOutputDir
+  def designTopModule = plan.getBitstreamPlanTopModule
+  def vendor = plan.getBitstreamPlanVendor
+  def board = plan.getBitstreamPlanBoard
+  def visibleFiles = plan.getBitstreamPlanVisibleFiles
+
+  def vivadoSynth =
+    def sourceTcl = source "fpga-shells/{vendor}/common/tcl/synth-only.tcl"
```
richardxia: Is this the only line that is different this function and `makeBitstream`? I would very strongly request that we factor out some internal function that has this as a parameter or at least some switch that allows you to select between the `synth-only.tcl` and the `vivado.tcl` file?

top.wake
```diff
@@ -24,6 +24,36 @@ global def compileSimTarget dut buildRoot =
 global topic vivadoVsrcHooks : (dut: DUT) => Option (Unit => List Path)
 global topic vivadoBlackBoxResources : DUT => Option (List String)
 
+global def fpgaSynthTarget board dut buildRoot =
```
richardxia, line 27: I have similar questions about this as for the other changes.

richardxia (commented):
Thanks for submitting this, but I think I would push more strongly for refactoring such that there is less code duplication. It's going to be difficult to keep these two in sync.

I can also take a stab at doing the refactor, to show what I would try.

fpga/mcs.wake
erikdanie: The issue with this, and why I made it completely separate, is that the only return value I care about is the error status of the vivado job. So I essentially want to run pretty much the normal vivado job (with a different top tcl file) but not do anything with the outputs. The reason is that I want to use this as a form of FPGA CI, but I want it to just run through the beginning of the process and see if it errors or not

erikdanie: And I don't know of a way to refactor the wake code such that a job can have varying return types, so I just copy-pasted

top.wake
```diff
@@ -24,7 +24,15 @@ global def compileSimTarget dut buildRoot =
...
+  setUpBitstream board dut buildRoot
+  | fpgaSynthOnly
+
 global def bitstreamTarget board dut buildRoot =
+  setUpBitstream board dut buildRoot
+  | makeBitstream
+
+global def setUpBitstream board dut buildRoot =
```
richardxia:
```suggestion
def setUpBitstream board dut buildRoot =
```

Can you make this a private function so that no one outside of this file can use it? It minimizes our API surface area.

fpga/mcs.wake
```diff
@@ -19,7 +19,17 @@ def vc707migHook =
...
+    runVivadoWithPlan plan "fpga-shells/xilinx/common/tcl/synth-only.tcl"
+
 global target makeBitstream plan =
+  def vivadoBitstream = runVivadoWithPlan plan  "fpga-shells/xilinx/common/tcl/vivado.tcl"
+  BitstreamOutputs
+  vivadoBitstream.getVivadoOutputBitstream
+  vivadoBitstream.getVivadoOutputAllOutputs
+  plan
+
+global target runVivadoWithPlan plan tclFile =
```
richardxia:
```suggestion
target runVivadoWithPlan plan tclFile =
```

fpga/mcs.wake
richardxia: Just wanted to open up my original request of renaming the function to be a verb, since it actually does something. Otherwise, this reads like a configuration option.

top.wake
erikdanie: Done, though I am not 100% sure that nothing outside this file uses this function as is

fpga/mcs.wake
erikdanie: Done

fpga/mcs.wake
erikdanie: Yes, sorry I forgot that. changed it to `runFPGASynthOnly`

richardxia (approved):
LGTM! Thanks for making my suggested changes!

Also, let me cc @mmjconolly so that he's aware of this PR, since this will interact with his gradual replacement of hardcoded paths to files in other repos, but this should merge soon, now.

PR #71, opened 2020-06-24
Use variable for rocket-chip location

mmjconolly:
(Also bumps rocket-chip in wit-manifest.json)

If rocket-chip were to change depth within the wit checkout layout, the consumers of rocket-chip would need to use this variable to find it.


richardxia (approved):
LGTM!

PR #78, opened 2020-07-06
Bump wit dependencies for wake 0.18 compatibility

richardxia:
/cc @nategraff-sifive, who cannot be assigned as a reviewer.

To facilitate the migration to Wake 0.18, I'm setting the versions of these dependencies to the earliest version which supported Wake 0.18. These should all be compatible with Wake 0.17.2. However, because part of api-generator-sifive's role is that of a metapackage, setting these versions here will force all dependents to at least require a version of these repositories which supports Wake 0.18.

I intentionally did not bump to the latest versions of all these repositories since I wanted to minimize the number of changes for dependents. My primary goal here is Wake 0.18-compatibility.


richardxia:
Rebased this on top of master so that I could have CI checks run.

PR #79, opened 2020-07-06
Add missing dependency on fpga-shells

mmjconolly:
`api-generator-sifive` has wake rules that expect wit to checkout `fpga-shells` but does not explicitly list it in the wit-manifest.
This was noticed when a wake rule in this repo was changed to use a symbol from fpga-shells


richardxia (approved):
LGTM!

PR #80, opened 2020-07-06
Add CI check to ensure that all Wit dependencies type check.

richardxia:
This PR adds a basic CI check in GitHub Actions to initialize the Wit workspace and check that Wake will pass type-checking.

We keep accidentally introducing little errors that could be caught by CI checks, but because we have none here, we only find out once we actually try to use them. It's also often difficult to tell if the direct wit dependencies of api-generator-sifive.git are actually correct, since we always use api-generator-sifive.git in some larger integration repository, where some other dependencies could bump the shared dependencies.

However, I'm hitting some weird Wit error in GitHub Actions that I don't exactly understand:

```
[ERROR] Two dependencies have the same name but an unrelated git history:
  git@github.com:sifive/api-generator-sifive.git
  /github/workspace/wit-packages/api-generator-sifive
```

It's really weird because api-generator-sifive is a top-level workspace package, but I'm wondering if this means that we have a cycle in our dependency tree. For what it's worth, I modeled this CI check after [api-firrtl-sifive's](https://github.com/sifive/api-firrtl-sifive/blob/ec61df3a2a83fd4653e55a1df865944fbff67c3c/.github/workflows/test.yml), which pulls the same trick of cloning the version of that repo into a `wit-packages` directory that lives outside of the wit workspace so that the `wit init` command can just point directly to it.

Here are some commands to reproduce it locally:

```bash
# Create temporary directory to be the parent of both wit-packages/ and the wit-workspace/
mkdir wit-test
cd wit-test
mkdir wit-packages

# Follow the exact same steps that GitHub Actions uses to clone api-generator-sifive
mkdir wit-packages/api-generator-sifive
cd wit-packages/api-generator-sifive
git init
git remote add origin https://github.com/sifive/api-generator-sifive
git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +d41c8fffa599123b7b8670e8c0b5fcd5ce3b31ea:refs/remotes/origin/add-ci
git checkout --progress --force -B add-ci refs/remotes/origin/add-ci
cd ../..

wit init wit-workspace -a ./wit-packages/api-generator-sifive -a git@github.com:sifive/environment-blockci-sifive.git::0.4.0
```

I've confirmed that this fails on both wit 0.12.0 and 0.13.0


mmjconolly:
```
+ exec wit init wit-workspace -a ./wit-packages/api-generator-sifive -a git@github.com:sifive/environment-blockci-sifive.git::0.4.0
Creating new workspace [/github/workspace/wit-workspace]
Cloned api-generator-sifive
The workspace now depends on 'api-generator-sifive::7c4d99d'
```
~I don't see commit 7c4d99d~

<edit> my bad it creates a merge commit earlier on

mmjconolly (approved):

richardxia:
Thanks for fixing this!

I also did a sanity check by pulling your changes and creating a temporary branch where I reverted your recent fix for the fpga-shell dependency and confirming that it flags the missing Wake variable as a problem: https://github.com/sifive/api-generator-sifive/runs/842911118

I'm going to merge this in, but let me send an FYI to several other people who sometimes contribute to this repo, since this means that we'll have CI set up now: @mwachs5, @albertchen-sifive, @rmac-sifive, @erikdanie, @nategraff-sifive, @RajeshVaradharajan

PR #81, opened 2020-07-11
Remove LocalOnly for VCS compilation.

richardxia:
This does not appear to be necessary anymore, possibly due to improvements to wake and fuse-wake.

I tested this locally on both the block-pio-sifive flow as well as some of our larger internal flows, and it seemed to work fine.

Note that I did attempt to do the same for Xcelium, but it seems that the FUSE sandbox is still interacting strangely with Xcelium.

@albertchen-sifive, do you have any recollection as to why we added this in the first place? I am going to assume that it was probably due to the Wake sandboxing not being robust back then, but just wanted to double check whether there was some specific edge case that it was failing on.


albertchen-sifive:
I think VCS has always worked fine with fuse, it's just that the output was not reproducible because `.fuse/123/*` paths would get baked into the binary. so while it works fine if you delete the output and rerun with exact same inputs you get hash mismatch errors

terpstra:
Ah. So you need the unshare change to make it pass `wake -c`.

richardxia:
Got it, thanks! Let me give https://github.com/sifive/wake/pull/444 a try to see if it can pass `wake -c`.

richardxia:
> Ah. So you need the unshare change to make it pass `wake -c`.

I realized/figured out that this is not true; with `wake -c`, you still get the same FUSE PID, so this would only be detectable if we ran multiple wake/fuse-waked processes.

But FWIW, `wake -c` does in fact pass on this PR without the `unshare` changes, so maybe this can be merged in anyway? I'm not sure if this was previously failing `wake -c` because we previously we spun up multiple fuse-waked process, but it at least does not seem to be the case today.

albertchen-sifive:
Even if it does work with `wake -c`, I think VCS still violates the `setPlanLocalOnly True` contract because its outputs depend on where it is running, which might mess with remote execution and stuff.

richardxia:
> Even if it does work with `wake -c`, I think VCS still violates the `setPlanLocalOnly True` contract because its outputs depend on where it is running, which might mess with remote execution and stuff.

What _is_ the contract for `setPlanLocalOnly True`? The code comments only say https://github.com/sifive/wake/blob/7aa9b140090ee868874bc773032b4e19edf741f9/share/wake/lib/system/job.wake#L88:

> Must run directly in the local workspace; no output detection performed

These days, I'm mostly treating `setPlanLocalOnly True` as a proxy for "Let FUSE/LD_PRELOAD attempt to sandbox filesystem operations". We/Wake are somewhat conflating "path-relocatable" with "sandboxing", since the only reason I want to remove `LocalOnly` is because I _want_ FUSE to do sandboxing and output file detection, even if I don't intend to run it remotely.

I'm also thinking that the only way we can really feasibly do remote execution when wrapping other tools is to do heavier sandboxing, such as using mount/filesystem namespaces. We will certainly want to run tools like VCS remotely, so we will have to do extra work to make them safe to run remotely.

mmjconolly:
> These days, I'm mostly treating `setPlanLocalOnly True` as a proxy for "Let FUSE/LD_PRELOAD attempt to sandbox filesystem operations". We/Wake are somewhat conflating "path-relocatable" with "sandboxing", since the only reason I want to remove `LocalOnly` is because I _want_ FUSE to do sandboxing and output file detection, even if I don't intend to run it remotely.

Just for call-site clarity, perhaps we could have a helper function added to wake stdlib named something very obvious like `def disableSandboxing plan` or `def enableSandboxing plan` which does whatever nessesary to turn off/on fuse/preload for the job.

richardxia:
> Just for call-site clarity, perhaps we could have a helper function added to wake stdlib named something very obvious like `def disableSandboxing plan` or `def enableSandboxing plan` which does whatever nessesary to turn off/on fuse/preload for the job.

Created a Wake ticket: https://github.com/sifive/wake/issues/446

mmjconolly (approved):

PR #82, opened 2020-07-14
Add support for coverage

richardxia:
This is still a WIP PR, and I may still decide to make changes to the global interfaces defined here.

That said, I think this should be a 100% backwards-compatible change with the existing codebase, as I've attempted to not change any public interfaces.

Two caveats with the current approach:

- `runSim`/`runSimWith` is not going to scale, as this needs to be yet another top-level target. I've instead created a `runSimV2`, which in theory could cover all the use cases of the other `run*Sim*` targets, only with a much more verbose CLI. I'll provide an example lower in this description, and I plan to actually add a README before considering this ready to merge.
- Anything that relies on `makeUserSimulationOptions` is not going to get the new coverage option. You must migrate to `makeUserSimulationOptionsV2`.

The new CLI is much more verbose, and here are some examples:

```sh
# Run VCS without waves:
wake -x 'runSimV2 pioDUT (makeSimTargetOptions "vcs" False)'

# Run Verilator with waves and coverage:
wake -x 'runSimV2 pioDUT (makeSimTargetOptions "vcs" True | setSimTargetOptionsCoverage True)'

# Run Xcelium without waves but with coverage:
wake -x 'runSimV2 pioDUT (makeSimTargetOptions "xcelium" False | setSimTargetOptionsCoverage True)'
```

TODOs:
- Get some people to test this as is
- Finish up the Xcelium integration. Also start and finish the Verilator integration
- Write a README
- Double check that these coverage reports are all actually passing
- Provide a simpler helper to replace `runSingleSim` that is compatible with `runSimV2`.


richardxia:
This is ready for a final review. I've addressed all of the TODOs I listed in my PR description, which was primarily about adding support for Xcelium and Verilator.

Note that Verilator does not respect the `UserSimulationOptions` `CoverageDir` or `CoverageTestName` options, but it seems that the only way to change this would be to modify verilator-main.cpp. By default, our implementation just writes the coverage data to the simulation execution directory, to `coverage.dat`, so it looks like by default it will avoid the issue of multiple test runs writing to the same directory that the proprietary tools have without any extra arguments anyway.

PR #86, opened 2020-08-11
Fix unused definitions and arguments

terpstra:
This PR corrects all the wake 0.19.0 warnings in this repository.


terpstra:
The two controversial changes were made as distinct commits.
- I removed the Test type entirely; it seems fully unused @albertchen-sifive 
- I removed some sort of vc707 hook @erikdanie

richardxia (approved):
Wow, this is great! Totally makes the unused variable check worth it.

erikdanie:
That vc707 hook stuff may need to be reintroduced, but i definitely think it belongs in fpga-shells regardless

PR #93, opened 2020-08-12
wit-manifest: bump for newer ldscript-generator

terpstra:
This is necessary to get the eccscrub symbol freedom-metal needs.


richardxia (approved):
LGTM

PR #96, opened 2020-09-26
Add new WAKE_IP_RESOURCE_LIBRARIES hook for vcs_wrapper.

richardxia:
This adds another API for passing files from an environment resource to VCS. This adds Verilog libraries, which are passed to VCS with the `-y` flag.


mmjconolly (approved):

richardxia:
Finally got around to testing this more, and I think this is good to merge now.

PR #100, opened 2020-10-08
Add VisibleFiles to DUTSimulator.

richardxia:
We did not previously have an API for adding files that should be visible (in the sense of Wake `Plan`s) to the simulator compiler, which meant that it was not possible to pass in Paths to files within library directories or include directories to the RTL simulators, short of passing in those files as source file arguments.

Example of where this is a problem:

- `main.sv` does a `` `include`` of `lib_a.sv`. `main.sv` is dynamically generated and may or may not actually include `lib_a.sv`
- `lib_a.sv` is located in a directory named `mylibs/`
- `lib_a.sv` may not always compile depending on factors related to when `main.sv` would include `lib_a.sv` in the first place

When you call `vcs +incdir+mylibs/ main.sv lib_a.sv`, it will be able to locate and include `lib_a.sv` and it will also compile fine. In the case where `lib_a.sv` should not be included/compiled, this will fail because `lib_a.sv` is not compilable, but `main.sv` didn't even want to include it in the first place. Normally this would not be a problem because you would just omit `lib_a.sv` from the command line, and VCS would only include and compile the files that are `` `include`d in the Verilog source.

When you call `vcs +incdir+mylibs/ main.sv` outside of Wake, it will normally be able to search for `lib_a.sv` in `mylibs/`, since this functions just like `-I` in `gcc`.

When you call `vcs +incdir+mylibs/ main.sv` from Wake but without `lib_a.sv` in the Wake visible file list, it won't be able to find and include the file.

Sorry for the not-so-great explanation above, but the bottom line is that we should have an API for allowing people to pass in directories of files that could be included, but should not necessarily be always be compiled.

What this translated to in code was adding a `VisibleFiles` property to `DutSimCompileOptions` and threading that through to the individual simulator compile rules. Note that Xcelium was special out of the three because it didn't already have a property for Visible files, so I had to add that to that simulator tuple as well.

I only tested this on VCS, but I think this is relatively safe to merge because it's the same change made for each simulator.

### DIFF
--- a/sim/dut-simulator.wake
+++ b/sim/dut-simulator.wake
@@ -124,9 +124,14 @@ tuple DUTSimCompileOptions =
   global SourceFiles:    List Path
   global Plusargs:       List NamedArg
   global Resources:      List String # resources to pass to the simulator `Plan`
+  # Files that should be visible to the simulator compiler but not passed in as
+  # direct source files. This should generally be used in conjuction with
+  # IncludeDirs and LibraryDirs in order to make files within those directories
+  # visible to the simulator compiler.
+  global VisibleFiles:   List Path
 
 global def makeDUTSimCompileOptions includeDirs defines sourceFiles plusargs =
-  DUTSimCompileOptions includeDirs Nil defines sourceFiles plusargs Nil
+  DUTSimCompileOptions includeDirs Nil defines sourceFiles plusargs Nil Nil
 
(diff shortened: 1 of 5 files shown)


mmjconolly (approved):
lgtm

gopitr (approved):
It works in sesame with vcs simulator.

PR #103, opened 2020-10-18
Adding filelist argument in vcs wrapper

vishvesh03:


scripts/vcs_wrapper
```diff
@@ -15,6 +15,11 @@ for file in $files; do
   file_args+=( "$file" )
 done
 
+filelists="$WAKE_IP_RESOURCE_FILELIST"
+filelistargs=( )
+for filelist in $filelists; do
+ filelistargs+=(-f "$filelist")
```
richardxia: Thanks for adding this. Could I make one small request here that instead of using the `-f` flag, we use the `-F` flag? `-F` is similar to `-f` except that it interprets path inside the filelist as relative to the path of the filelist. E.g. `vcs -F /path/to/filelist.F` will prepend `/path/to` to each path inside of `filelist.F`. This allows us to more easily change the paths to the files in the future if we preserve the relative path between the `.F` file and the actual Verilog files.

vishvesh03: Agree, will change to `-F`

scripts/vcs_wrapper
```diff
@@ -15,6 +15,11 @@ for file in $files; do
   file_args+=( "$file" )
 done
 
+filelists="$WAKE_IP_RESOURCE_FILELIST"
+filelistargs=( )
```
richardxia:
```suggestion
filelist_args=( )
```

For consistency with all of the other variables, can you put an underscore before the word "args" in this variable?

scripts/vcs_wrapper
```diff
@@ -15,6 +15,11 @@ for file in $files; do
...
+for filelist in $filelists; do
+ filelistargs+=(-F "$filelist")
```
richardxia:
```suggestion
 filelist_args+=(-F "$filelist")
```

scripts/vcs_wrapper
```diff
@@ -35,4 +40,5 @@ done
 
 IFS="$OIFS"
 
-vcs "${file_args[@]}" "${include_args[@]}" "${library_args[@]}" "${static_lib_args[@]}" "${pli_tab_args[@]}" "$@"
+vcs "${file_args[@]}" "${include_args[@]}" "${filelistargs[@]}" "${library_args[@]}" "${static_lib_args[@]}" "${pli_tab_args[@]}" "$@"
```
richardxia:
```suggestion
vcs "${file_args[@]}" "${include_args[@]}" "${filelist_args[@]}" "${library_args[@]}" "${static_lib_args[@]}" "${pli_tab_args[@]}" "$@"
```

scripts/vcs_wrapper
```diff
@@ -35,4 +40,5 @@ done
...
+
```
richardxia: Could you remove this extra blank line you added to the file? In general, we should avoid adding superfluous whitespace in PRs.

scripts/vcs_wrapper
```diff
@@ -15,6 +15,11 @@ for file in $files; do
   file_args+=( "$file" )
 done
 
+filelists="$WAKE_IP_RESOURCE_FILELIST"
```
richardxia:
```suggestion
filelists="$WAKE_IP_RESOURCE_FILELISTS"
```

Can you make this environment variable plural as to match the other environment variables?

richardxia (commented):
Thanks, I have a few specific comments about the code itself, but this looks mostly good to me.

Am I correct to assume that you have tested this in an environment where this new environment variable is being set? I would like to make sure that we've at least done a manual test of this before merging it in.

scripts/vcs_wrapper
vishvesh03: I have tested in soc-sesame environment with the above changes, the build seems clean to me.

scripts/vcs_wrapper
vishvesh03: Yeah sure

richardxia (approved):
LGTM! Thanks!

PR #104, opened 2020-10-21
syntax error fixed in vcs_wrapper

vishvesh03:


richardxia (approved):
LGTM.

In the future, could you double check that any extra changes that you make to a branch still work? I'm assuming that this bug was introduced with some of the extra changes that I had requested.

vishvesh03:
@richardxia i have pushed with syntax error previously, can you merge this again, I have rectified the syntax error

vishvesh03:
> LGTM.
> 
> In the future, could you double check that any extra changes that you make to a branch still work? I'm assuming that this bug was introduced with some of the extra changes that I had requested.

Actually, this bug got introduced while merging with master, Sorry about that.
will make sure about this in future

